PR #2, opened 2021-12-08
feat(config): allow to edit the underlying config

maxklenk:


src/types/internal.types.ts
```diff
@@ -44,3 +45,15 @@ export type ExecuteCrossParams = {
 export type UpdateStep = (step: Step, execution: Execution) => void
 export type UpdateExecution = (execution: Execution) => void
 export type CallbackFunction = (updatedRoute: Route) => void
+
+export type Config = {
+  apiUrl: string
+  rpcs: Record<ChainId, string[]>
+  multicallAddresses: Record<ChainId, string | undefined>
```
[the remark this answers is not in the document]
Addminus: These should be the multicall addresses the sdk can use to query the balances for a wallet address.

src/Lifi.ts
```diff
@@ -13,24 +14,43 @@ import {
...
+    return this.config
   }
 
   getPossibilities = async (
     request?: PossibilitiesRequest
   ): Promise<PossibilitiesResponse> => {
+    if (!request) request = {}
+
+    // apply defaults
+    request.bridges = request.bridges || this.config.defaultRouteOptions.bridges
```
mathiasmoeller, line 49: shouldn't this be the intersection of the requested bridges from `request` and the bridges that where enabled in the `defaultRouteOptions`?

src/config.ts
```diff
@@ -0,0 +1,82 @@
...
+  config.defaultExecutionSettings = Object.assign(
+    config.defaultExecutionSettings,
+    configUpdate.defaultExecutionSettings
+  )
+
+  // OPTIONS
+  config.defaultRouteOptions = Object.assign(
+    config.defaultRouteOptions,
+    configUpdate.defaultRouteOptions
+  )
```
mathiasmoeller, line 79: `Object.assign` doesn't do a deep comparison, right? So this will always overwrite the whole `defaultRouteOptions` even if the user only changed a single value inside. Do we want this behaviour or do we want a deep merge?

src/executionFiles/bridges/anyswap.execute.ts
```diff
@@ -130,8 +135,8 @@ export class AnySwapExecutionManager {
     waitForTxProcess.txLink =
       toChain.metamask.blockExplorerUrls[0] + 'tx/' + waitForTxProcess.txHash
     waitForTxProcess.message = 'Funds Received:'
-    status.fromAmount = parsedReceipt.fromAmount
-    status.toAmount = parsedReceipt.toAmount
+    // status.fromAmount = parsedReceipt.fromAmount
+    // status.toAmount = parsedReceipt.toAmount
```
mathiasmoeller, line 139: Was this removed on purpose? Please document why this is commented out 😇

src/config.ts
```diff
@@ -0,0 +1,82 @@
...
+  Hooks,
+  multicallAddresses,
+} from './types'
+
+function chainIdToObject<T>(val: T): Record<ChainId, T> {
+  const result: Record<number, T> = {}
+
+  const values = Object.values(ChainId)
+  values.forEach((chainId) => {
+    if (typeof chainId !== 'string') {
```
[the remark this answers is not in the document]
maxklenk: That's what I would expect as well, but I needed to add this check to only use the values and not the keys too. For some reason both are included.

src/Lifi.ts
maxklenk: It is easy to create a blocking combination if allowed and denied bridges. I thought it may be clearer for the user that if he defines such an object it is actually used. And not merged with another one internally.

src/executionFiles/bridges/anyswap.execute.ts
maxklenk: Oh sorry, the "comment" is only in the commit. There is actually no real parsing of the receipt build for anyswap yet, so these amounts are always `'0'` which makes following requests fail.

src/Lifi.ts
mathiasmoeller: Hmm I agree but also strange if you define something in the configuration step that doesn't have any effect in the next step

src/config.ts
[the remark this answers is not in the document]
mathiasmoeller: Okay 😅

src/executionFiles/bridges/anyswap.execute.ts
mathiasmoeller: Ah alright 👍

src/config.ts
maxklenk: It does not do a deep merge, but it should merge every top level attribute on its own and not overwrite the whole object.

PR #4, opened 2021-12-10
feat(nxtp): handle encryption/decryption via hooks

mathiasmoeller:


src/executionFiles/bridges/nxtp.execute.ts
```diff
@@ -69,10 +82,7 @@ export class NXTPExecutionManager {
       if (!this.shouldContinue) return status
       // -> request key
       try {
-        const encryptionPublicKey = await (window as any).ethereum.request({
-          method: 'eth_getEncryptionPublicKey',
-          params: [await signer.getAddress()], // you must have access to the specified account
-        })
+        const encryptionPublicKey = await hooks.getPublicKeyHook()
```
maxklenk, line 85: Before executing the whole encryption process, we should check if a `getPublicKeyHook` is defined or not. Because if there is no hook we can skip the whole process and also not show it in the UI. The default `getPublicKeyHook` has to be undefined to make this work

src/executionFiles/bridges/nxtp.execute.ts
```diff
@@ -201,22 +215,38 @@ export class NXTPExecutionManager {
...
+      preparedTransaction.txData.receivingChainId,
+      preparedTransaction.txData.receivingChainTxManagerAddress,
+      signer
     )
 
     claimProcess.status = 'PENDING'
     claimProcess.message = 'Waiting for claim'
     update(status)
 
+    const callData = await this.decryptData(
```
maxklenk, line 236: It is great that we are now in charge of handling the decryption. We should update the claimProcess, if the user has to decrypt the data by setting the message and status right.

maxklenk: if no decrypt hook was defined by the user we don't need to update the status.

src/executionFiles/bridges/nxtp.execute.ts
```diff
@@ -240,4 +270,46 @@ export class NXTPExecutionManager {
...
+  private decryptData = async (
+    decryptHook: DecryptHook,
+    { txData, encryptedCallData }: ReceiverTransactionPreparedPayload
+  ): Promise<string> => {
+    let callData = '0x'
+    if (txData.callDataHash !== utils.keccak256(callData)) {
+      try {
+        callData = await decryptHook(encryptedCallData)
+      } catch (e) {
+        // TODO: update process failed
```
maxklenk, line 309: We may not want to catch the error here and instead handle it outside

maxklenk (changes requested):
Works awesome.

src/executionFiles/bridges/nxtp.execute.ts
mathiasmoeller: Ah yeah forgot about that todo.
Don't we want to set the process to failed here?
That's also how we do it above

maxklenk (approved):

PR #5, opened 2021-12-10
refactor!: refactor status handling

Addminus:
At the moment the function that dictates what is done in order to update a step with new execution states is defined in Lifi.ts. The goal is to move the update logic for execution changes to the status.ts file and to improve the readability of related code in general.

BREAKING CHANGE: interface of external status management methods changed


src/Lifi.ts
```diff
@@ -152,14 +148,12 @@ class LIFI {
...
-        stepExecutor = new StepExecutor()
-        this.activeRoutes[route.id].executors.push(stepExecutor)
-        await stepExecutor.executeStep(
-          signer,
-          step,
-          updateFunction,
-          this.activeRoutes[route.id].settings.switchChainHook
+        stepExecutor = new StepExecutor(
+          route,
+          this.activeRoutes[route.id].settings
```
mathiasmoeller, line 153: Whats the difference between `route` and `this.activeRoutes[route.id]`?

src/Lifi.ts
```diff
@@ -152,14 +148,12 @@ class LIFI {
...
         )
+        this.activeRoutes[route.id].executors.push(stepExecutor)
```
mathiasmoeller, line 155: Saving the Route inside the StepExecuter and the StepExecuter inside the Route could create some kind of recursive problems? 
Or is that the reason why there is a difference between `route` and `this.activeRoutes[route.id]`? If yes that's pretty implicit and hard to grasp

mathiasmoeller: Maybe we just store the executor in a map of routeIds?

src/executionFiles/StepExecutor.ts
```diff
@@ -30,6 +34,15 @@ export class StepExecutor {
 
   executionStopped = false
 
+  constructor(
+    route: Route,
+    settings: EnforcedObjectProperties<ExecutionSettings>
+  ) {
+    this.route = route
```
mathiasmoeller, line 41: Why do we need to store the route here? I can't find where it is being used

src/executionFiles/bridges/horizon.execute.ts
```diff
@@ -125,133 +119,188 @@ export class HorizonExecutionManager {
...
-          update(status)
+          updateExecution(currentExecution)
           const operation = await bridgeSDK.api.getOperation(operationId)
           console.debug('operation', operation)
 
           // Send > Wait
           if (
             operation.actions[0].status === 'in_progress' &&
-            allowanceAndCrossProcess.status === 'ACTION_REQUIRED'
+            allowanceAndCrossProcess.currentExecution === 'ACTION_REQUIRED'
```
mathiasmoeller, line 133: Wait, is this refactoring on purpose? The `status` inside the `process` should still be called `status`, right? Only the old `status` object (used in `update(status)`) should be changed. The status inside the process is a string containing the current status and should be still called like that I think

src/status.ts
```diff
@@ -2,87 +2,139 @@
 /* eslint-disable @typescript-eslint/ban-types */
```
mathiasmoeller, line 2: Maybe rename the file

src/status.ts
```diff
@@ -2,87 +2,139 @@
...
+   * @param  {Step} step  The current step in execution
+   * @return {Execution, UpdateExecution}       The initialized execution object for this step and a function to update this step
+   */
+  initExecutionObject = (step: Step) => {
+    const currentExecution =
+      step.execution || (deepClone(emptyExecution) as Execution)
+    const updateExecution = (newExecution: Execution) => {
+      step.execution = newExecution
+      this.settings.updateCallback(this.route)
+    }
```
mathiasmoeller, line 44: This updates the step, but the route returned is still the old one. Does this only work because the step has a reference to the route? If yes we need to make this explicit. If a developer deep clones the step on his side this could break I guess.

src/status.ts
```diff
@@ -2,87 +2,139 @@
...
-  params?: object
-): void => {
-  status.status = 'FAILED'
-  currentProcess.status = 'FAILED'
-  currentProcess.failedAt = Date.now()
-  if (params) {
-    for (const [key, value] of Object.entries(params)) {
-      currentProcess[key] = value
+  /**
+   * Reduces a sequence of names to initials.
```
mathiasmoeller, line 52: What? 😁

src/executionFiles/bridges/horizon.execute.ts
mathiasmoeller: Pretty sure this breaks smth 😅

maxklenk (changes requested):
Awesome, the change look very good: clean code, documentation, good structure/order of functions.

I just marked some small things, also there are not tests, it may be easy to add unit tests for the `StatusMananger` to allow easier refactoring later.

src/StatusManager.ts
```diff
@@ -0,0 +1,210 @@
...
+   * @param  {string} processId  The process id to update
+   * @param  {Status} status The status the process gets.
+   * @param  {object} [params]   Additional parameters to append to the process.
+   * @return {Process} The update process
+   */
+  updateProcess = (
+    step: Step,
+    processId: string,
+    status: Status,
+    params?: object
```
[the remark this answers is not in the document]
Addminus: Good point! Will fix!

src/StatusManager.ts
```diff
@@ -0,0 +1,210 @@
+/* eslint-disable max-params */
+/* eslint-disable @typescript-eslint/ban-types */
```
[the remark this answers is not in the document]
Addminus: I will look into that

src/StatusManager.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    if (!currentProcess) {
+      throw new Error('Cannot find process for given id')
+    }
+
+    switch (status) {
+      case 'CANCELLED':
+        currentProcess.doneAt = Date.now()
+        currentProcess.message =
+          'CANCELLED - Funds have been refunded on source chain.'
```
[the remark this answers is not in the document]
mathiasmoeller: I would leave it as it is for now since we only have this one usecase. 
As soon as we localize the frontend we will have to remove all those messages and replace them with codes anyways. Then the frontend can decide what to render for a `CANCELLED` status depending on the context. And it also has the information about the chain etc.

src/StatusManager.ts
```diff
@@ -72,7 +72,7 @@ export default class StatusManager {
    * @return {Step}       The step with the updated execution object
    */
   updateExecution(step: Step, status: Status, receipt?: Receipt): Step {
-    if (!step.execution) throw Error('Can update empty execution')
+    if (!step.execution) throw Error("Can't update empty execution")
```
Addminus, line 75: lol🤣

Addminus (commented):
LGTM. Can't approve, because I'm the author.

@maxklenk ?

PR #7, opened 2021-12-15
fix(balance): split large token lists into chunks

mathiasmoeller:
Multicalls can only take 100 calls in one request. Therefore, larger token lists need to be split up into chunks﻿


src/balances/utils.int.spec.ts
```diff
@@ -78,5 +79,13 @@ describe('balances utils', () => {
       const tokens: Token[] = []
       await loadAndCompareTokenAmounts(walletAddress, tokens)
     })
+
+    it('should handle large lists', async () => {
```
Addminus, line 83: maybe specify what a large list is? 
in this case it's more than 100 entries. Everything below that would work just fine.

'should handle lists with more then 100 entries'

Addminus (changes requested):
Great Job!

Addminus (approved):

PR #13, opened 2022-01-03
Lf-140: docstrings for user facing functions

Addminus:


src/Lifi.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
+/* eslint-disable max-len */
```
mathiasmoeller, line 2: Eslint allows to define different rules for different types of lines. We should rather just increase/disable the max-length for comments instead of disabling it completely.

https://eslint.org/docs/rules/max-len

src/Lifi.ts
```diff
@@ -28,18 +29,31 @@ import {
...
+   * @return {Config} - The config object
+   */
+  getConfig = (): Config => {
     return this.config
   }
 
-  setConfig = (configUpdate: ConfigUpdate) => {
+  /**
+   * Set a new confuration for the SDK
+   * @param {ConfigUpdate} configUpdate - The new configuration. Same as interface Config but with optional values.
```
mathiasmoeller, line 44: Maybe make it shorter and simpler? 
```suggestion
   * @param {ConfigUpdate} configUpdate - The configuration fields that should be updated.
```
If not then it should at least be "real" sentences.

src/Lifi.ts
```diff
@@ -59,6 +73,11 @@ class LIFI {
     return result.data
   }
 
+  /**
+   * Get a set of routes for a request that describes a transfer of coins.
```
mathiasmoeller, line 77:
```suggestion
   * Get a set of routes for a request that describes a transfer of tokens.
```

src/Lifi.ts
```diff
@@ -59,6 +73,11 @@ class LIFI {
...
+   * @param {RoutesRequest} routesRequest - A description of the transfer.
+   * @return {Promise<RoutesResponse>} The resulting routes that can be used to realize the described transfer of coins.
```
mathiasmoeller, line 79:
```suggestion
   * @return {Promise<RoutesResponse>} The resulting routes that can be used to realize the described transfer of tokens.
```

src/Lifi.ts
```diff
@@ -103,13 +132,24 @@ class LIFI {
...
     if (!this.activeRoutes[route.id]) return
     for (const executor of this.activeRoutes[route.id].executors) {
       executor.stopStepExecution()
     }
   }
 
+  /**
+   * Execute a route.
+   * @param {Signer} signer - The signer required to send the transactions.
+   * @param {Route} route - The route that is to be executed. Cannot be an active route.
```
mathiasmoeller, line 149: K.I.S.S.
```suggestion
   * @param {Route} route - The route that should be executed. Cannot be an active route.
```

src/Lifi.ts
```diff
@@ -121,6 +161,13 @@ class LIFI {
     return this.executeSteps(signer, route, settings)
   }
 
+  /**
+   * Resume the execution of a route that has been stopped or had an error while executing.
```
mathiasmoeller, line 165:
```suggestion
   * Resume the execution of a route that has been stopped or had an error during the previous execution.
```

src/Lifi.ts
```diff
@@ -213,15 +265,29 @@ class LIFI {
     }
   }
 
+  /**
+   * Get a list of active routes.
```
mathiasmoeller, line 269: It is not just any list of active routes, it is THE list of active routes
```suggestion
   * Get the active routes.
```

src/Lifi.ts
```diff
@@ -213,15 +265,29 @@ class LIFI {
...
+   * @return {Route[]} A list of routes.
+   */
   getActiveRoutes = (): Route[] => {
     return Object.values(this.activeRoutes).map((dict) => dict.route)
   }
 
+  /**
+   * Get a single updated route. The route has to be active.
```
mathiasmoeller, line 277:
```suggestion
   * Return the current route information for given route id
```

src/Lifi.ts
```diff
@@ -213,15 +265,29 @@ class LIFI {
...
+   * @param {Route} route - A route object.
+   * @return {ReturnValueDataTypeHere} Brief description of the returning value here.
```
mathiasmoeller, line 279: Please add a "Brief description of the returning value here"

src/Lifi.ts
```diff
@@ -213,15 +265,29 @@ class LIFI {
...
+   */
   getActiveRoute = (route: Route): Route | undefined => {
     return this.activeRoutes[route.id].route
   }
 
-  // Balances
+  /**
+   * Returns the balances of a specific token a wallet holds across all chains we have aggergated. .
```
mathiasmoeller, line 286:
```suggestion
   * Returns the balances of a specific token a wallet holds across all chains we have aggregated.
```

src/Lifi.ts
```diff
@@ -213,15 +265,29 @@ class LIFI {
...
+   * @param {string} walletAddress - A walletaddress.
```
mathiasmoeller, line 287: Do we have a naming convention here already? I think "wallet address" is a bit more common?

src/Lifi.ts
```diff
@@ -237,6 +303,12 @@ class LIFI {
     return balances.getTokenBalance(walletAddress, token)
   }
 
+  /**
+   * Returns the balances for a list tokens a wallet holds across all chains we have aggergated.
```
mathiasmoeller, line 307:
```suggestion
   * Returns the balances for a list tokens a wallet holds across all chains we have aggregated.
```

mathiasmoeller: Maybe "we" is also a bit to informal here. "Across all aggregated chains" maybe?

src/Lifi.ts
```diff
@@ -252,6 +324,12 @@ class LIFI {
     return balances.getTokenBalances(walletAddress, tokens)
   }
 
+  /**
+   * This method queries the balances of tokens for a specific list of chains for a given wallet. .
```
mathiasmoeller, line 328:
```suggestion
   * Queries the balances of tokens for a specific list of chains for a given wallet.
```

src/helpers.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable max-len */
```
mathiasmoeller, line 1: Same here

src/Lifi.ts
mathiasmoeller: in other places here we used "wallet address"

mathiasmoeller (commented):
Those are just suggestions ;)

src/Lifi.ts
Addminus: ohh my bad

src/Lifi.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
+/*eslint max-len: ["error", { "ignoreComments": true }]*/
```
mathiasmoeller, line 2: I would rather do that in the global eslint configuration? We don't need maxlength errors on comments anywhere I would say

Addminus: good point, completely forgot that we have one...

src/Lifi.ts
```diff
@@ -237,6 +302,12 @@ class LIFI {
     return balances.getTokenBalance(walletAddress, token)
   }
 
+  /**
+   * Returns the balances for a list tokens a wallet holds  across all aggregated chains.
+   * @param {string} walletAddress - A walletaddress.
```
mathiasmoeller, line 307:
```suggestion
   * @param {string} walletAddress - A wallet address.
```

mathiasmoeller (changes requested):
One last :P

mathiasmoeller (approved):

PR #17, opened 2022-01-05
LF-196: parse receipts of bridges

maxklenk:


src/executionFiles/bridges/anyswap.ts
```diff
@@ -155,35 +157,60 @@ const getTransferStatus = async (txHash: string) => {
...
+    'event LogAnySwapIn(bytes32 indexed txhash, address indexed token, address indexed to, uint amount, uint fromChainID, uint toChainID)',
+  ]
+  const interfaceSwap = new ethers.utils.Interface(abiSwap)
+  receipt.logs.forEach((log) => {
+    try {
+      const parsed = interfaceSwap.parseLog(log)
+      const amount = parsed.args.amount as BigNumber
+      result.toAmount = amount.toString()
+    } catch (e) {
+      // find right log by trying to parse them
```
mathiasmoeller, line 192: Does that mean the only way to find out if that receipt log belongs to a swap or a transfer is by trying to parse it with the correct ABI and if it succeeds it is of this type?

src/executionFiles/bridges/anyswap.ts
```diff
@@ -155,35 +157,60 @@ const getTransferStatus = async (txHash: string) => {
...
+    try {
+      const parsed = interfaceTransfer.parseLog(log)
+      if (parsed.args['to'].toLowerCase() === toAddress) {
+        result.toAmount = parsed.args['value'].toString()
+        result.toTokenAddress = log.address.toLowerCase()
+      }
+    } catch (e) {
+      // find right log by trying to parse them
+    }
+  })
```
mathiasmoeller, line 212: This part confuses me a bit. We don't know which log is the correct one so we try to parse all of them?
Why not use only the fallback if that works for all ERC20 tokens?

maxklenk: The bridges/exchange sometimes have custom events which we can parse to see what has happened. Since they often do not contain everything we need we also check for tokenTransfer events or changes in native tokens if needed. 
The Transfer Events of ERC20 tokens are triggered by the token contracts, if a contract does not have them implemented it is also good to have the custom events as fallback.

I have now separated them better.

src/executionFiles/utils.ts
```diff
@@ -0,0 +1,214 @@
...
+    if (tx.blockNumber && providerBlockNumber) {
+      const balanceBefore = await provider.getBalance(
+        toAddress,
+        tx.blockNumber - providerBlockNumber - 1
+      )
+      const balanceAfter = await provider.getBalance(
+        toAddress,
+        tx.blockNumber - providerBlockNumber
+      )
+      const balanceDiff = balanceAfter.sub(balanceBefore)
```
mathiasmoeller, line 106: Is this really necessary? Shouldn't we be able to read the balance from the transaction?
Getting the balance for other than the latest block number was very slow when I tried it

mathiasmoeller: I see that this is only called when the `toAmount` is 0. But why does that happen?

src/executionFiles/utils.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  // gas
+  result = {
+    ...result,
+    ...getUsedGas({ tx, receipt }),
+  }
+  return result
+}
+
+export const getUsedGas = (params: {
```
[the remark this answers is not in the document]
maxklenk: I exported them because it may make sense to used dem individually, eg. to change the order. And when exported we can test them.

src/executionFiles/utils.ts
maxklenk: The `toAmount` is the value we want to find, we usually know what token we expect to get and have an estimate how much, but to get the exact amount we parse the logs.
This function only has to be used when it was not possible to find any other transferred tokens. In this case the user got the native token which does not trigger an event when transferred, instead an internal transaction transfers these tokens.

The only solution I came up to read this from the chain is checking if the gas balance before and after the specific transaction has changed.

src/executionFiles/utils.ts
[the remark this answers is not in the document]
mathiasmoeller: But we don't use them right now and we don't have tests for them right now 😋

mathiasmoeller (approved):

PR #22, opened 2022-01-13
feat: improve handling of wallet & backend errors

mathiasmoeller:
* Create custom error types for easier handling
* Improve the parsing of wallet and Lifi Backend related errors.

### DIFF
--- a/src/Lifi.ts
+++ b/src/Lifi.ts
@@ -25,6 +25,8 @@ import {
   ExecutionSettings,
 } from './types'
 import StatusManager from './StatusManager'
+import { parseBackendError } from './utils/parseError'
+import { ValidationError } from './utils/errors'
 
 class LIFI {
   private activeRouteDictionary: ActiveRouteDictionary = {}
@@ -52,6 +54,7 @@ class LIFI {
    * Get a set of current possibilities based on a request that specifies which chains, exchanges and bridges are preferred or unwanted.
    * @param {PossibilitiesRequest} request - Object defining preferences regarding chain, exchanges and bridges
    * @return {Promise<PossibilitiesResponse>} Object listing current possibilities for any-to-any cross-chain-swaps based on the provided preferences.
+   * @throws {LifiError} Throws a LifiError if request fails.
    */
   getPossibilities = async (
     request?: PossibilitiesRequest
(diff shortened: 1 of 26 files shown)


mathiasmoeller:
To see how the changes are handled on the frontend: https://github.com/lifinance/lifi-web/pull/177

src/Lifi.ts
```diff
@@ -64,22 +67,26 @@ class LIFI {
...
   /**
    * Get a set of routes for a request that describes a transfer of tokens.
    * @param {RoutesRequest} routesRequest - A description of the transfer.
    * @return {Promise<RoutesResponse>} The resulting routes that can be used to realize the described transfer of tokens.
+   * @throws {LifiError} Throws a LifiError if request fails.
    */
   getRoutes = async (routesRequest: RoutesRequest): Promise<RoutesResponse> => {
     if (!isRoutesRequest(routesRequest)) {
-      throw new Error('SDK Validation: Invalid Routs Request')
+      throw new ValidationError('Invalid Routs Request')
```
Addminus, line 89: weee wooo typo alarm!

src/utils/parseError.ts
```diff
@@ -0,0 +1,157 @@
...
+    ? ` (${formatTokenAmountOnly(
+        step.action.fromToken,
+        step.action.fromAmount
+      )} ${step.action.fromToken.symbol} on ${
+        getChainById(step.action.fromChainId).name
+      })`
+    : ''
+
+  transactionNotSend +=
+    ", please retry.<br/>If it still doesn't work, it is safe to delete this transfer and start a new one."
```
Addminus, line 67: I don't think this is a good idea. We should stay away from html in the sdk as it is also can be used on the server or any cmd line script.

<br> should be \n
The link should be rendered in the frontend

Addminus: IMO the text  for the process message is complete and holds enough info up to this point.
("[...] to delete this transfer and start a new one."). 

The sentence "You can check the failed transaction _here_" should be a frontend rendered thing.

Addminus (changes requested):

src/utils/parseError.ts
mathiasmoeller: We need the html rendering here since we want to include the link to the transaction explorer which can not be done w/o HTML. 
It is not an issue if this is used on a server since this is just additional context information. The developer doesn't need to parse it, it is merely a comfort feature. 

I don't have a strong opinion on this but max suggested to move it to the sdk since it has more information about what exactly went wrong and where to look for solutions.

Addminus (approved):

PR #24, opened 2022-01-18
fix(Hop): check subgraphs for success confirmation instead of using hop sdk

mathiasmoeller:
So far we used `waitForDestinationChainReceipt` to track when a hop transfer has been fulfilled.
The method checks the chain for new transaction. If the transaction has been executed earlier already it will not be returned from this function.

To solve this problem 
1. we check the hop subgraph of the source chain for the transfer containing the transaction id. This will give us the transfer id.
2. Then we check the subgraph of the destination chain for a withdraw with the given transfer id. This will give us the transaction hash on the destination chain.
3. Finally we can load the transaction receipt using the transaction hash.

Problem for L1->L2 transfers: On ETH we have `transferId === transactionId`. That allows us to skip the `getTransferIdOnSourceChain` call. However, transfers from ETH do not appear in the L2 subgraphs (yet?).
Neither with `withdrawalBondeds` (that's how it usually works), nor with `transferFromL1Completeds` (which sounds exaclty like what we need)
For those reasons we will use the Hop SDK for this scenario.


src/executionFiles/bridges/hop.ts
```diff
@@ -3,117 +3,101 @@ import {
...
 
-const getHopBridge = (bridgeCoin: CoinKey) => {
-  isInitialized()
-  if (!Object.keys(bridges).length) {
-    throw Error(
-      'No HopBridge available! Initialize Hop implementation first via init(signer: JsonRpcSigner, chainId: number, toChainId: number)'
-    )
-  }
-  return bridges[bridgeCoin]
+const getTransferIdByTxHash = async (
```
maxklenk, line 37: Perhaps the function name show contain that we only search for transfers on the sending side.

src/executionFiles/bridges/hop.ts
```diff
@@ -3,117 +3,101 @@ import {
...
-  fromChainId: number,
-  toChainId: number
-) => {
-  isInitialized()
-  const bridge = getHopBridge(bridgeCoin)
-  const hopFromChain = hopChains[fromChainId]
-  const hopToChain = hopChains[toChainId]
-  const tx = await bridge.approveAndSend(amount, hopFromChain, hopToChain)
-  return tx
+const getTxHashByTransferId = async (
```
maxklenk, line 59: The function name should tell that this only works for withdraws.

maxklenk (changes requested):
Nice. Pretty cool to get rid of a dependency. On the other hand we now fully rely on working subgraphs. As we have seen with connext they are not as reliable as the chain itself. Maybe this is good for now, but in the future - e.g. when moved to the backend, I would prefer to have a fallback.

src/executionFiles/bridges/hop.ts
mathiasmoeller: Good point

src/executionFiles/bridges/hop.ts
mathiasmoeller: The function allows to search for transfers on both sides, depending on which chain you pass, that's why I gave it a general name. 
But one could argue that it doesn't make so much sense to look on the receiving side 😅  Ill give both functions a similar naming

maxklenk (approved):

PR #25, opened 2022-01-21
LF-241: fix(hop): find transactions for L1 to L2 transfers

maxklenk:
When sending transfers using a bridge we want to find out when it was actually finished on the receiving chain. This PR tracks transfers from L1 (Ethereum) to L2 (any other chain).

Issue: https://lifi.atlassian.net/browse/LF-241

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "@connext/nxtp-sdk": "^0.1.8",
     "@ethersproject/abi": "^5.5.0",
     "@ethersproject/contracts": "^5.4.0",
-    "@hop-protocol/sdk": "^0.0.1-beta.286",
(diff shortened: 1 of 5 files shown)


Addminus (changes requested):
Pretty cool not to having to rely on the hop-sdk anymore and being able to get older transfers.

One thing: When reading through the code, I tripped over `waitForDestinationChainReceiptL2` multiple times. I thought it meant waiting for a receipt on an L2 chain, but that does not seem to be the only usecase for this function. Therefore, I would rename it to `waitForDestinationChainReceiptL2ToX`. Just a bit longer but in line with `waitForDestinationChainReceiptL1toL2`.

Addminus (approved):
Sweet!

PR #26, opened 2022-01-21
refactor: check if allowance script can access metamask wallet

Addminus:
RP for this frontend PR: https://github.com/lifinance/lifi-web/pull/184

If this is merged before Frontend, the frontend behaviour should not change


mathiasmoeller (approved):
LGTM

PR #27, opened 2022-01-24
fix: resolve eslint issues

mathiasmoeller:
This does not solve all eslint issues but it reduces the number from 88 to 35. 
There are some things that have to be discussed but since it touches so many files I want to get this merged ASAP.


src/executionFiles/utils.ts
```diff
@@ -152,7 +163,7 @@ export const getTransferredNativeTokenBalanceFromLog = (params: {
     try {
       const parsed = interfaceGas.parseLog(log)
       const result = {
-        tokenAddress: log.address.toLowerCase(),
+        toTokenAddress: log.address.toLowerCase(),
```
mathiasmoeller, line 166: Please take extra care here, I think this was wrong before but I could be wrong.

mathiasmoeller:
Uff why the hell do those errors happen in the node_modules? 🙄

src/executionFiles/utils.ts
maxklenk: Oh yes, that should be toTokenAddress. This function is not used/needed a lot, that's why we may have missed it.

maxklenk (approved):

PR #33, opened 2022-01-31
fix: check chain before sending transactions

mathiasmoeller:
Switching chains while preparing transaction (`Lifi.getStepTransaction`) causes an unexpected failure when trying to send the transaction (`signer.sendTransaction`) because the signer is configured for the wrong chain.
To prevent that we need to check here again that the user is on the correct chain.

### DIFF
--- a/src/executionFiles/StepExecutor.ts
+++ b/src/executionFiles/StepExecutor.ts
@@ -5,7 +5,6 @@ import {
   BridgeTool,
   CrossStep,
   ExchangeTool,
-  getChainById,
   Hooks,
   LifiStep,
   Step,
@@ -21,6 +20,7 @@ import { openocean } from './exchanges/openocean'
 import { paraswap } from './exchanges/paraswap'
 import { SwapExecutionManager } from './exchanges/swap.execute'
 import { uniswap } from './exchanges/uniswaps'
+import { switchChain } from './switchChain'
 
 export class StepExecutor {
   settings: Hooks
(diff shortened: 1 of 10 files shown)


jasonliudan (approved):
It looks good for me !

src/executionFiles/StepExecutor.ts
```diff
@@ -52,53 +52,28 @@ export class StepExecutor {
...
-        }
-      } catch (e: any) {
-        this.statusManager.updateProcess(step, switchProcess.id, 'FAILED', {
-          errorMessage: e.message,
-          errorCode: e.code,
-        })
-        this.statusManager.updateExecution(step, 'FAILED')
-        throw e
-      }
+    const updatedSigner = await switchChain(
```
Addminus, line 55: shouldn't that only occur in the stepexecutors, when its really needed?

src/executionFiles/allowance.execute.ts
```diff
@@ -38,14 +38,14 @@ export const checkAllowance = async (
 
       if (new BigNumber(amount).gt(approved)) {
```
Addminus, line 39: IF you remove the chaincheck in the `StepExecutor.ts`, don't forget to add it here.

src/executionFiles/switchChain.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @param step
+ * @param switchChainHook
+ * @param allowUserInteraction
+ */
+export const switchChain = async (
+  signer: Signer,
+  statusManager: StatusManager,
+  step: Step,
+  switchChainHook: SwitchChainHook,
+  allowUserInteraction: boolean
```
Addminus, line 22: maybe default this to `false`?

Addminus (changes requested):
I tested the code manuallz to check if it would work with the frontend "auto-resume-feature-on-page-load".

The code in itself works great. Chain switches and everything work while the modal is opened. However if i reload the page, the execution status is not updated with a switch chain status message when it's necessary. If i open the modal, the execution resumes instantly. 
It would be great to inform the user about the required chain switch.

src/executionFiles/StepExecutor.ts
Addminus: this might be scope creep, as it is another issue: https://lifi.atlassian.net/browse/LF-126

src/executionFiles/allowance.execute.ts
Addminus: this might be scope creep, as it is another issue: https://lifi.atlassian.net/browse/LF-126

src/executionFiles/StepExecutor.ts
mathiasmoeller: Yeah I would agree that thats a different feature. My implementation should not change the behavior here, just extract the code. Let's deal with that problem in the other ticket

src/executionFiles/switchChain.ts
mathiasmoeller: Why? it's not an optional value and right now we don't have a place where we don't want to pass a value to it. So it will always be defined

PR #34, opened 2022-01-31
feat: code style and eslint improvements

chybisov:
The first iteration of the unification of code style and `eslint` rules for our repositories. This will help current team members and newcomers write more consistent code. After discussion and approving these changes, we can create a [submodule](https://github.blog/2016-02-01-working-with-submodules/) repo and link it with all others.

A couple of majors:
1.	Adds proven and well maintained `eslint-config-airbnb` which takes care of best practices and helps avoid many code-smells. https://github.com/airbnb/javascript
2.	Try always to use named exports. As the team grows, we will ensure every person uses the same import names as other team members to avoid confusion and increase future readability. Another reason is that if the user doesn't need all the functionality we provide, unused code could be removed via tree-shaking, resulting in a smaller bundle size. Always a good practice for building libraries. Readings https://github.com/rollup/rollup/issues/1961#issuecomment-423037881 https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad
3.	Return of semi. In general, most of the repos from great teams and React ecosystem, like React, Mui, Storybook, prettier, Airbnb, Stripe, etc., use them and feel good. Small conviction booster https://levelup.gitconnected.com/to-semicolon-or-not-to-semicolon-a2e337827bcb
4.	Adds `curly` and `comma-dangle` rules. Examples of incorrect code:
```
if (foo) foo++;

while (bar)
    baz();

if (foo) {
    baz();
} else qux();
```
Examples of correct code:
```
if (foo) {
    foo++;
}

while (bar) {
    baz();
}

if (foo) {
    baz();
} else {
    qux();
}
```
5.	Adds `.gitattributes` with `* text=auto eol=lf` to avoid line endings collisions between different OS.

This repo currently contains **133 problems (99 errors, 34 warnings)** that need to be addressed or disabled (open topic for discussion).

Please review and give feedback. 😃


maxklenk:
Very good, I would love to align the rules across the projects and update them to our code even cleaner. The new rules contain a lot of useful rules and I think more rules lead to an error instead of a warning which is good to ensure that we take care of them.

My personal feedback about some settings:

- `no-use-before-define` => I like to read code from the top down, a function may call sub-functions further down. On the other hand the advantage of the rule would be that it is handled the same in all files - which would be good

- `no-await-in-loop` => How do we want to handle this? We have a few cases where we currently use/need await in loops. Because we do want to do one execution after another in the loop. We could disable the rule in using a disable comment, so developers have to think about it twice.

- `semicolons` => We don't use them so far, I like this. It reduces the code. You have to write and read less. We are using typescript which is very good in avoiding the few cases where ASI may alter the code. There is the `no-unexpected-multiline` ESLint rule which should help to find all cases in question.

- `submodule` => How would this setup with keeping these settings in a submodule look like? Would it be possible to do the same using a npm package? I would prefer a npm package to avoid dealing with submodules.

mathiasmoeller:
Nice 👍 
`no-await-in-loop`: Never really understood what the issue here is. There are some valid cases where you want to loop over something and make sure the code is executed after each other, not in parallel. 
`semicolons`: I don't have a strong opinion on that, I am fine with whatever you guys decide

Regarding the submodule with the configurations, I guess an npm package would only allow us to have the eslint configuration for example as a eslint plugin, but not the .gitattributes, prettier config etc? Or how would be do that. Do we configure eslint etc to use the configuration files placed in the `node_modules` then? But I guess for a `submodule` we would also need to adjust the paths to the configuration files since they usually are placed in the project root

Addminus (commented):
- `semicolons`: I've taken on the habit of not using them. I actually like the code better that way and I like the solution @maxklenk suggests here. 
- `no-await-in-loop`: This would definitively need some refactoring in our codebase. Is there really the need for this?
- `no-use-before-define`:  Don't mind this one
- ` submodules`: Don't mind this one

maxklenk:
enable `no-use-before-define`
enable `no-await-in-loop` => document how and why to disable it
enable `import/no-default-export` => document where to use const ui = {}
no plan to use index files => overhead, less advantage
error as unknown, function to parse e to error with code and message
```ts
const isError = function (exception: any): exception is Error {
  return typeof exception == 'object'
    && exception !== null
    && 'name' in exception && typeof exception.name === 'string'
    && 'message' in exception && typeof exception.message === 'string'
}

try {
} catch (e) {
  if (isError(e)) {
    console.log(e.message)
  }
}
```


=> make a pool about `semicolons`

=> open question how to test internal functions without exposing everything

PR #38, opened 2022-02-08
build: add action to release beta versions

mathiasmoeller:
This action will create and publish beta releases to Github and NPM.
* It listens to pushes to the `main` branch
* It bumps the version number to a beta version (e.g. `0.4.4-beta.0`)
* It updates the changelog 
image
* Those changes are not persisted in the branch. That means that in normal releases the beta information will not be in the changelog, which is intended

### DIFF
--- a/.github/workflows/beta-release.yml
+++ b/.github/workflows/beta-release.yml
@@ -1,0 +1,59 @@
+name: Release & Publish
+
+on:
+  push:
+    paths-ignore:
+      - 'docs/**'
+      - 'README.md'
+      - 'LICENSE'
+      - '.editorconfig'
+    branches:
+      - main
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/beta-release.yml
```diff
@@ -0,0 +1,59 @@
+name: Release & Publish
+
+on:
```
maxklenk, line 3: How is checked that this only runs for beta releases and not for real releases?

.github/workflows/beta-release.yml
```diff
@@ -0,0 +1,59 @@
...
+  push:
+    paths-ignore:
+      - 'docs/**'
+      - 'README.md'
+      - 'LICENSE'
+      - '.editorconfig'
+    branches:
+      - main
```
maxklenk, line 11: How did you test that the action does what you want?
We may add this feature branch to the list to do a test-beta-release and can then remove the branch again.

maxklenk (changes requested):

maxklenk (approved):

PR #41, opened 2022-02-18
fix: handle tokens with 0xeee... zero addresses

Addminus:


src/balances/utils.ts
```diff
@@ -133,7 +133,10 @@ const executeMulticall = async (
   // Collect calls we want to make
   const calls: Array<Call> = []
   tokens.map((token) => {
-    if (token.address === constants.AddressZero) {
+    if (
+      token.address === constants.AddressZero ||
+      token.address == '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee'
```
chybisov, line 138: Can we move this to constants as well?

chybisov (changes requested):

src/balances/utils.ts
Addminus: Changed it: Added a function for zero address checks

chybisov: Great!

PR #43, opened 2022-02-24
feat: use common bridge executor for all known BridgeTools

mathiasmoeller:
All our bridges now support the /status endpoint and can be handled with the general handler. 
This also implicitly adds Hyphen bridge support :)


Addminus (approved):
nice and tidy!

PR #47, opened 2022-03-21
feat: add infinite approval to execution settings

mathiasmoeller:
- feat: add infinite approval to execution settings
- test: disable failing int test


mathiasmoeller:
Alright it seems that more receipt tests fail here than locally. Will ignore them for now since we will move the whole receipt parsing to the BE we can simply remove all the code then (https://lifi.atlassian.net/browse/LF-442)

PR #52, opened 2022-04-05
fix: route undefined bug

Addminus:


mathiasmoeller:
> It passes a flag to the StatusManager that controls wether it should update the state of a route or not. We have two methods that stop a route execution:
> 
> * `moveExecutionToBackground()` - this sends a stop flag to all executors which will then stop all execution when they encounter a stopping flag in the executors. This might still result in status changes on the route as only state that require user Interactions have a stopping flag before them.
> * `stopExecution` - stops execution and deletes all references to the active routes. StepExecutioner-wise it still uses all the same logic. This is where the error comes from: no reference but updates to the route still happen because the executioners continue to the next stopping flag. Introducing a stopping flag to the StatusManager was the easiest to prevent that. Now we can soft stop(stop executioners but allow for updates with non user interaction states) and hard stop(stop executioners and stop StatusManagers)

cool, thanks for the explanation! :)

mathiasmoeller (approved):

PR #60, opened 2022-04-22
fix: remove unnecessary chain switch

Addminus:


src/executionFiles/bridges/bridge.execute.ts
```diff
@@ -45,6 +63,7 @@ export class BridgeExecutionManager {
     if (!oldCrossProcess || !oldCrossProcess.txHash) {
       if (action.fromToken.address !== constants.AddressZero) {
         // Check Token Approval only if fromToken is not the native token => no approval needed in that case
+        await chainSwitchClosure()
```
mathiasmoeller, line 66: This will not stop execution here if `!updatedSigner`. Is this on purpose?

Addminus: the chainswitch does that

mathiasmoeller: How? It simply returns the execution, but nothing is done with that information as far as I can see?

Addminus:
In fact the goal of this PR is not achievable. Closing it.

PR #61, opened 2022-05-04
feat: added getTools

H3xept:


mathiasmoeller (approved):
Apart from that small test fix (the other tests won't run otherwise) it looks good 👍 
Feel free to merge it after you fixed it.
We can wait with the SDK release until we also added the /tokens methods.

PR #62, opened 2022-05-04
fix!: deep clone incoming routes to prevent side effects

mathiasmoeller:
**ATTENTION**: This might be a breaking change if people were relying on the side effects (like our FE did)


mathiasmoeller:
Let's not merge this until @chybisov tested that it solves the issues in the widget

chybisov (approved):

PR #63, opened 2022-05-04
feat!: improve status management

chybisov:
This PR allows us to create proper process message handling (e.g. i18n) in the widget while preserving old messages for the SDK users that need them.

1. Defined process types
2. Made process messages depending on type and status
4. Error messages cleanup
5. Added curly eslint rule
6. Organized files and folders a little bit

Depends on https://github.com/lifinance/types/pull/55

### DIFF
--- a/src/execution/StatusManager.ts
+++ b/src/execution/StatusManager.ts
@@ -21,7 +21,6 @@ interface Receipt {
 type InternalUpdateRouteCallback = (route: Route) => void
 
 interface OptionalParameters {
-  // message?: ProcessMessage
   txHash?: string
   txLink?: string
   doneAt?: number
@@ -29,9 +28,6 @@ interface OptionalParameters {
   errorMessage?: string
   htmlErrorMessage?: string
   errorCode?: string | number
-
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  // [key: string]: any
 }
 
 /**
--- a/src/execution/StatusManager.ts
+++ b/src/execution/StatusManager.ts
@@ -3,14 +3,14 @@ import {
   Execution,
   InternalExecutionSettings,
   Process,
-  ProcessMessage,
   ProcessType,
   Route,
   Status,
   Step,
   Token,
-} from './types'
-import { deepClone } from './utils/utils'
+} from '../types'
+import { deepClone } from '../utils/utils'
+import { getProcessMessage } from './utils'
 
 interface Receipt {
   fromAmount?: string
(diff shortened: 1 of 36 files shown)


src/execution/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+    return message.replace(`{${index}}`, arg)
+  }, initialMessage)
+}
+
+const processMessages: Record<
+  ProcessType,
+  Partial<Record<Status, string | ((args?: string[]) => string)>>
+> = {
+  TOKEN_ALLOWANCE: {
+    STARTED: 'Setting token allowance.',
```
Addminus, line 69: `ACTION_REQUIRED` is missing here

src/execution/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+    PENDING: 'Waiting for token allowance approval.',
+    DONE: 'Token allowance approved.',
+  },
+  SWITCH_CHAIN: {
+    PENDING: 'Chain switch required.',
```
Addminus, line 74: And here we use `CHAIN_SWITCH_REQUIRED`

src/execution/StatusManager.ts
```diff
@@ -20,15 +21,17 @@ interface Receipt {
 type InternalUpdateRouteCallback = (route: Route) => void
 
 interface OptionalParameters {
-  message?: ProcessMessage
+  // message?: ProcessMessage
```
Addminus, line 24: can this be removed?

src/execution/StatusManager.ts
```diff
@@ -20,15 +21,17 @@ interface Receipt {
...
+  txLink?: string
   doneAt?: number
   failedAt?: number
   errorMessage?: string
   htmlErrorMessage?: string
   errorCode?: string | number
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  [key: string]: any
+  // [key: string]: any
```
Addminus, line 34: this as well

src/execution/StatusManager.ts
```diff
@@ -90,39 +95,32 @@ export default class StatusManager {
...
       startedAt: Date.now(),
-      message: message,
-      status: 'PENDING',
-    }
-    if (params) {
-      for (const [key, value] of Object.entries(params)) {
-        newProcess[key] = value
-      }
+      message: getProcessMessage(type, status ?? 'STARTED'),
+      status: status ?? 'STARTED',
```
Addminus, line 123: You are defaulting to `STARTED` if no status is given. But in `processMessages` not every type has a `STARTED` value. Wouldn't that be a problem?

Addminus (changes requested):

Addminus:
We should mark this as breaking change.

PR #71, opened 2022-05-24
LF-321: update sdk with acceptSlippageUpdateHook

Addminus:


src/types/internal.types.ts
```diff
@@ -64,6 +65,14 @@ export type SwitchChainHook = (
...
 ) => Promise<Signer | undefined>
 
+export type AcceptStepUpdateHook = (
+  oldReturnAmount: string,
+  newReturnAmount: string,
+  returnCurrency: Token,
+  oldSlippage: number,
+  newSlippage: number
+) => Promise<boolean | undefined>
+
```
chybisov, line 75: Does it make sense to make params here as an object? Something like 

```
enum StepUpdateType {
  Slippage = 'slippage',
  ...
}

interface StepUpdateBase {
  type: StepUpdateType;
}

interface SlippageStepUpdate extends StepUpdateBase {
  token: Token;
  oldAmount: string;
  newAmount: string;
  oldSlippage: number;
  newSlippage: number;
}

interface OtherStepUpdate extends StepUpdateBase {
  ...
}

type StepUpdate = SlippageStepUpdate | OtherStepUpdate;

export type AcceptStepUpdateHook = (update: StepUpdate) => Promise<boolean | undefined>
```

So if we need to use this hook for some other update in the future, it'll be extensible.

chybisov (changes requested):

src/types/internal.types.ts
Addminus: ohh missunderstood your review. Extracting the method params like this does make more sense than my approach. Introducing the StepUpdateType as a base, however, does not reaaly give us that much. Updating the step possibly updates every "receiving values" as one. We cannot change what we update. Also i don't see the value in giving the opportunity to give those possibilities.. The user has defined the route. The only updates necessary are the recalculation of the routes when preparing the transaction data. 

This new hook is to warn the user if the return values of the transaction have changed drastically (and are outside of the set slippage limit) between the initial calculation of the route and the execution.

chybisov (approved):

PR #76, opened 2022-06-01
feat: use tryBlockAndAggregate to fetch balances

mathiasmoeller:
My idea was to isolate broken tokens by chunking down to calls of 1 to make sure we get all the other tokens.
Not sure if the log is the best idea but I didn't see another good way how to inform about the broken token.

### DIFF
--- a/src/utils/multicall.ts
+++ b/src/utils/multicall.ts
@@ -1,43 +1,11 @@
 import { Fragment, Interface, JsonFragment } from '@ethersproject/abi'
 import { Contract } from '@ethersproject/contracts'
 import { getRpcProvider } from '../connectors'
-import { Bytes } from 'ethers'
+import { BigNumber, Bytes } from 'ethers'
 import { splitListIntoChunks } from './utils'
+import MULTICALL_ABI from './multicallAbi.json'
 
 const MAX_MULTICALL_SIZE = 100
-
-export const MULTICALL_ABI = [
-  {
-    constant: true,
-    inputs: [
-      {
-        components: [
-          { name: 'target', type: 'address' },
-          { name: 'callData', type: 'bytes' },
-        ],
-        name: 'calls',
-        type: 'tuple[]',
-      },
-    ],
-    name: 'aggregate',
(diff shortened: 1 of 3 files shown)


src/balances/index.ts
```diff
@@ -5,6 +5,7 @@ const getTokenBalance = async (
   walletAddress: string,
   token: Token
 ): Promise<TokenAmount | null> => {
+  debugger
```
H3xept: Do we really want this?

src/utils/multicall.ts
```diff
@@ -90,10 +81,56 @@ export const fetchDataUsingMulticall = async (
...
+          )
+          return []
+        } else {
+          // try with chunk size of one to identify exactly the failed call
+          const smallestChunks = await chunkDataAndExecuteMulticall(
+            abiInterface,
+            multicallContract,
+            calls,
+            chainId,
+            1
```
H3xept, line 98: Correct me if I'm wrong but if a large chunk fails, this code will re-chunk that chunk with chunk size == 1 and re execute? 

If so, would it maybe be better to recurse by calling chunkSize / 2 ? This guarantees to reach the failing token in O(log(n)) calls and we won't waste bandwith / API call allowance / time in doing a ton of size 1 multicall calls.

H3xept:
I left my comments before seeing max's. He's right but I think tryAggregate does not allow us to find out which token is problematic?

Addminus:
I think we need both multicall`v2` and a solution for `v1`. I don't think `v2` is deployed to all chains we need/support

src/balances/index.ts
mathiasmoeller: This was removed loooong before your review :P

H3xept:
> I think it does, it returns a list of responses where the failing on is empty, so we can filter for that one and find the matching address.

Ok yeah the multicall contract returns a list of results which can also be not successful https://github.com/makerdao/multicall/blob/master/src/Multicall2.sol#L64

maxklenk (approved):

PR #77, opened 2022-06-07
fix: updated e2e + added tsc compilation on push

H3xept:


package.json
```diff
@@ -55,7 +56,10 @@
     "axios": "^0.27.2",
     "bignumber.js": "^9.0.1",
     "eth-rpc-errors": "^4.0.3",
-    "ethers": "^5.6.6"
+    "ethers": "^5.6.6",
+    "i": "^0.3.7",
+    "npm": "^8.12.1",
+    "qs": "^6.10.5"
```
chybisov, line 62: Are we sure we want these as dependencies?

chybisov (changes requested):

package.json
H3xept: Nope, I needed these for changes I ended up removing.

chybisov (changes requested):

package.json
```diff
@@ -65,6 +66,7 @@
     "@types/hdkey": "^2.0.0",
     "@types/jest": "^27.5.1",
     "@types/pino": "^7.0.5",
+    "@types/qs": "^6.9.7",
```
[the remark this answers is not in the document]
H3xept: eheh yes

chybisov (approved):

PR #79, opened 2022-06-28
feat: handle slippage errors

mathiasmoeller:
image

So far we never actually handled the case that between a /routes request and the according /getStepTransaction the quote changed more than the defined slippage.
We need proper handling in the FE to ask the user if they are fine with the new (bad) quote or if they want to abort.
This is on the roadmap.
As a quick solution we decided that the Backend throws an SlippageError if the slippage is 2x the defined slippage. We decided for this large amount to not cause to many failures until we have the real update logic in place. The problem is that this could also happen in the second or third step of a route where the funds of the user already moved a bit. That's why we didn't want to be too restrictive right now.
The current implementation will show the user the following error if the slippage goes wild (screenshot).

### DIFF
--- a/src/utils/errors.ts
+++ b/src/utils/errors.ts
@@ -6,6 +6,7 @@ enum ErrorType {
   ValidationError = 'ValidationError',
   NotFoundError = 'NotFoundError',
   UnknownError = 'UnknownError',
+  SlippageError = 'SlippageError',
 }
 
 export enum LifiErrorCode {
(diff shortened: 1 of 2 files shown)


src/utils/errors.ts
```diff
@@ -60,6 +62,9 @@ export class LifiError extends Error {
   ) {
     super(message)
 
+    // Set the prototype explicitly: https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
```
H3xept, line 65: 🥲

PR #81, opened 2022-07-05
feat: use new substates from api

Addminus:


src/execution/utils.ts
```diff
@@ -89,6 +100,32 @@ const processMessages: Record<ProcessType, Partial<Record<Status, string>>> = {
   },
   TRANSACTION: {},
 }
+const substatusMessages: Record<
+  StatusMessage,
+  Partial<Record<Substatus, string>>
+> = {
+  PENDING: {
+    BRIDGE_NOT_AVAILABLE: 'Bridge communication is temporarily unavailable',
+    CHAIN_NOT_AVAILABLE: 'RPC communication is temporarily not available',
```
chybisov, line 109: Let's stick to consistent wording here: `is temporarily unavailable.` or `is temporarily not available.` for both.

chybisov (approved):

PR #83, opened 2022-07-07
feat: new error messages

Addminus:


src/execution/exchanges/swap.execute.ts
```diff
@@ -167,7 +167,7 @@ export class SwapExecutionManager {
         error: {
           code: LifiErrorCode.TransactionFailed,
           message: 'Failed while waiting for receiving chain.',
-          htmlMessage: getTransactionFailedMessage(swapProcess),
+          htmlMessage: getTransactionFailedMessage(step, swapProcess.txHash),
```
mathiasmoeller, line 170: you are passing the `txHash` here but the method expects a `txLink`. is that on purpose?

src/utils/parseError.ts
```diff
@@ -150,6 +158,14 @@ export const parseError = async (
         e.stack
       )
     }
+
+    if (e.Code === LifiErrorCode.ValidationError) {
```
mathiasmoeller, line 162: Not sure if it makes a difference but it should probably be `e.code`

src/execution/exchanges/swap.execute.ts
Addminus: that was a mistake

src/utils/parseError.ts
Addminus: that does make a difference. I just copied the code from the transaction unprepared check. There it is `e.Code`. So we are getting parsed messages for this error now too :)

src/execution/bridges/bridge.execute.ts
```diff
@@ -165,7 +165,10 @@ export class BridgeExecutionManager {
         error: {
           code: LifiErrorCode.TransactionFailed,
           message: 'Failed while waiting for receiving chain.',
-          htmlMessage: getTransactionFailedMessage(crossChainProcess),
+          htmlMessage: getTransactionFailedMessage(
+            step,
+            crossChainProcess.txHash
```
mathiasmoeller, line 170: 🙃

src/utils/parseError.ts
```diff
@@ -142,14 +150,22 @@ export const parseError = async (
       )
     }
 
-    if (e.Code === LifiErrorCode.TransactionUnprepared) {
+    if (e.code === LifiErrorCode.TransactionUnprepared) {
```
mathiasmoeller, line 153: Are you sure that this one here isn't supposed to be `Code`? It's not one of our errors, could be `Code`

Addminus: yes i am pretty sure
image


And it is one of our errors

PR #86, opened 2022-07-14
fix: handle more multicall failure cases

maxklenk:
Catches parsing errors to avoid making the whole call fail.


maxklenk:
I disagree with the expected condition in the failing test. If there are valid tokens we for sure want to return their balance. For the failing ones we can just set balance `0` as we know that there can't be any funds.

PR #100, opened 2022-08-10
feat: skip step tx generation if tx data already provided

Addminus:


chybisov (approved):

src/execution/bridges/bridge.execute.ts
```diff
@@ -71,20 +71,21 @@ export class BridgeExecutionManager {
...
+          step = {
+            ...(await stepComparison(
+              statusManager,
+              personalizedStep,
+              updatedStep,
+              settings.acceptSlippageUpdateHook,
+              this.shouldContinue
+            )),
+            execution: step.execution,
+          }
```
mathiasmoeller, line 88: Are we sure this doesn't lead to problems if a normal step is executed, execution fails, it's retried and now contains already the old tx data but new should be fetched since it's a regular tx? Are we sure this cannot happen since we pass everything by reference?

Addminus: good catch! this might actually be true. It would be a good idea to set the transaction request to undefined in case of a restart of a failed transaction.

PR #102, opened 2022-08-22
feat: add version check

chybisov:
Sometimes our partners forget to update our SDK and other packages, which leads to unnecessary questions in the support channels etc., and we can avoid some of them. Let's notify them about a new SDK version in the console (Widget will also benefit from reusing this function).

image


Addminus (approved):
Good idea! Maybe show the current version to highlight how far behind they might be?

chybisov:
> Good idea! Maybe show the current version to highlight how far behind they might be?

image

Like this?

Addminus:
yeah that should do!

PR #110, opened 2022-09-14
feat: add bridgeExplorerLink to receiving chain process

chybisov:


src/execution/utils.ts
```diff
@@ -48,6 +48,7 @@ export async function waitForReceivingTransaction(
                 statusResponse.status,
                 statusResponse.substatus
               ),
+            txLink: statusResponse.bridgeExplorerLink,
```
mathiasmoeller, line 51: Maybe only update this if the link is actually there? Not sure if this could otherwise overwrite the previous "normal" `txLink`? Didn't look into the implementation of the statusManager though

chybisov: @mathiasmoeller there is no other link for receiving chain process at that time, it appears later when the bridging is done. Let's add this check to be saved in the future 😄

mathiasmoeller (approved):

PR #111, opened 2022-11-18
feat: add acceptExchangeRateUpdateHook and return gas info during route execution

chybisov:


src/utils/parseError.ts
```diff
@@ -95,6 +95,14 @@ export const getTransactionFailedMessage = (
     : baseString
 }
 
+export const getExchangeRateChangedMessage = (step: Step) => {
```
Addminus, line 98: With the new hook, this error message feels off. Especially with the premise, that the exchangerate hook is not really covering excessive slippage anymore. it just uses it as a threshold.

Maybe something like "Transaction was not sent, your funds are still in your wallet. The exchange rate has changed and the previous estimation can not be fulfilled due to severe value loss."

Addminus: This would cover cases when no hook is set, but also explain why the transaction failed when the user declined

Addminus (changes requested):
Looks good. I have a question regarding the exchange rate error message

src/utils/parseError.ts
chybisov: @Addminus done. I removed 'severe' though because it can be only 0.1% sometimes 😄

PR #114, opened 2023-02-08
refactor: replace deepClone with native structuredClone

chybisov:
Native [structuredClone()](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) 👀

CI will be fixed once #113 is merged.


0xksure (approved):
LGTM

PR #115, opened 2023-02-16
chore: adding quote handlers to sdk

Abhikumar98:
Adding function handlers to execute and manage quotes directly without creating routes from them.

This will majorly be useful for integrations when they generate quotes and this will give them direct option to execute it with management similar to route.


src/Lifi.ts
```diff
@@ -254,6 +254,23 @@ export default class LIFI {
...
+   * @return {Route} The stopped route.
+   */
+  stopQuoteExecution = (step: Step): Route => {
+    const currRoute = convertStepToRoute(step)
+
+    if (!currRoute) {
+      throw new ValidationError(
+        "Missing 'fromAmountUSD' | 'toAmountUSD' in step estimate."
+      )
+    }
```
chybisov, line 269: We throw from `convertStepToRoute` if the validation does not pass, so this piece is unreachable and should be unnecessary. 👀

src/Lifi.ts
```diff
@@ -254,6 +254,23 @@ export default class LIFI {
     return route
   }
 
+  /**
+   * Stops the execution of an active quote.
+   * @param {Step} step - A route that is currently in execution.
+   * @return {Route} The stopped route.
+   */
+  stopQuoteExecution = (step: Step): Route => {
+    const currRoute = convertStepToRoute(step)
```
chybisov, line 263: `currentRoute`? 🙂

src/Lifi.ts
```diff
@@ -299,6 +316,24 @@ export default class LIFI {
...
+   */
+  updateQuoteExecution = (
+    step: Step,
+    settings: Pick<ExecutionSettings, 'executeInBackground'>
+  ): void => {
+    const route = convertStepToRoute(step)
+
+    if (!route) {
+      return
+    }
```
chybisov, line 332: Same as above.

src/Lifi.ts
```diff
@@ -324,6 +359,63 @@ export default class LIFI {
...
+  executeQuote = async (
+    signer: Signer,
+    step: Step,
+    settings?: ExecutionSettings
+  ): Promise<Route> => {
+    const route = convertStepToRoute(step)
+
+    if (!route) {
+      return Promise.reject()
+    }
```
chybisov, line 380: Same as above.

src/Lifi.ts
```diff
@@ -324,6 +359,63 @@ export default class LIFI {
...
+  resumeQuote = async (
+    signer: Signer,
+    step: Step,
+    settings?: ExecutionSettings
+  ): Promise<Route> => {
+    const route = convertStepToRoute(step)
+
+    if (!route) {
+      return Promise.reject()
+    }
```
chybisov, line 403: Same as above.

src/Lifi.ts
```diff
@@ -324,6 +359,63 @@ export default class LIFI {
...
+  /**
+   --
+   return this.executeSteps(signer, clonedRoute, settings)
+   new route.steps = [ step ]
+   pass in this.executeRoute | this.executeQuote
+   --
+   write a resumeQuote function to execute the quote
+   --
+   maybe a function that accepts step and returns a route
+   */
```
chybisov, line 417: Do we still need these comments?

src/helpers.ts
```diff
@@ -91,3 +92,35 @@ export const checkPackageUpdates = async (
     // Cannot verify version, might be network error etc. We don't bother showing anything in that case
   }
 }
+
+export const convertStepToRoute = (step: Step): Route | null => {
```
chybisov, line 96: This will never return `null`.

chybisov (changes requested):

src/Lifi.ts
Abhikumar98: Nope, removing them.

src/Lifi.ts
Abhikumar98: cool, I'll avoid abbreviations. haha

src/helpers.ts
Abhikumar98: I'm trying to put a validation check on `estimate.toAmountUSD` and `estimate.fromAmountUSD` which is optional but required in Route.
So, if this field is `undefined` then throw an error.

and since this function is expecting something to be returned, I've returned null in case of error.

Hence, the 2 step validation.

Would it be recommended that if these two values are undefined, we take it as 0?

src/Lifi.ts
Abhikumar98: Linked to this comment: https://github.com/lifinance/sdk/pull/115#discussion_r1109572456

src/helpers.ts
chybisov: > and since this function is expecting something to be returned, I've returned null in case of error.

@Abhikumar98 It doesn't work like that 🙂  Please check the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw) 🙂 

> The throw statement throws a user-defined exception. Execution of the current function will stop (the statements after throw won't be executed), and control will be passed to the first [catch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch) block in the call stack. If no catch block exists among caller functions, the program will terminate.

Abhikumar98: Yes, this worked now.
surprisingly it was showing an error earlier about expecting a value to be returned.

Thanks @chybisov

chybisov (changes requested):
The code looks good. 

I was thinking about why do we need this and come up with a question 🙂

How will we use resume/stop/update methods for quotes after starting the execution? After we start the execution, we have the route object returned, and I see no reason to use quote methods anymore here, we can use route methods.

@Addminus

Abhikumar98:
> The code looks good.
> 
> I was thinking about why do we need this and come up with a question 🙂
> 
> How will we use resume/stop/update methods for quotes after starting the execution? After we start the execution, we have the route object returned, and I see no reason to use quote methods anymore here, we can use route methods.
> 
> @Addminus

This change is majorly around being able to directly execute the quotes, and it does makes sense that since route handlers are available we can get rid of the wrappers around it for quote specifically.

So, we can have a handler for executing Quote and then the returned Route can be handled the regular way?

Wdyt @Addminus @chybisov

chybisov:
@Abhikumar98 yes, the returned route can be handled with existing functionality. Maybe we can make `executeRoute` function accept the route and quote, and we can convert the quote to a route inside that function if necessary 🙂

Abhikumar98:
> @Abhikumar98 yes, the returned route can be handled with existing functionality. Maybe we can make `executeRoute` function accept the route and quote, and we can convert the quote to a route inside that function if necessary 🙂

We've decided to expose only the conversion function allowing the user to convert step to route and operate via existing route handlers rather than having duplicate logic written in places.

chybisov (approved):

Addminus:
Dang my message did not get send for some reason. But its essentially what abhishek said. We will only provide a function to convert quotes to routes.

when converting the quote to a route during the execution, a user would have to handle a type change during the execution flow. That seemed overly complicated to us.

PR #119, opened 2023-02-23
Remove references of getChainById exposed by types package

Abhikumar98:
Replacing all references of `getChainById` exposed by types package with instance of `ChainService.getChainById`.

LF-1186

### DIFF
--- a/src/utils/parseError.ts
+++ b/src/utils/parseError.ts
@@ -81,14 +81,15 @@ export const getTransactionNotSentMessage = async (
   return transactionNotSend
 }
 
-export const getTransactionFailedMessage = (
+export const getTransactionFailedMessage = async (
   step: Step,
   txLink?: string
-): string => {
+): Promise<string> => {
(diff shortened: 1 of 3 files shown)


src/utils/parseError.ts
```diff
@@ -81,14 +81,15 @@ export const getTransactionNotSentMessage = async (
...
 }
 
-export const getTransactionFailedMessage = (
+export const getTransactionFailedMessage = async (
   step: Step,
   txLink?: string
-): string => {
+): Promise<string> => {
+  const chainsService = ChainsService.getInstance()
+  const chainName = await chainsService.getChainById(step.action.toChainId)
```
chybisov, line 89: Should be `chain` 🙂

chybisov (changes requested):

src/utils/parseError.ts
Abhikumar98: updated ✅

chybisov (approved):

PR #121, opened 2023-03-03
fix: storing the timestamp of wallet approvals

Abhikumar98:
# Issue

There are process involving external user authorisation and the time spent in approval of those requests shouldn't be considered in the Swap/Bridge estimated time.

# Fix

Calculate the estimated pending time after omitting time taken for external approvals.

SDK will now hold the timestamp ( `approvedAt` ) at which the user approved the request

For a process involving the External action:

- `startAt` is the time in which the wallet confirmation/signature is triggered
- `approvedAt` is the time when the user approved the request

By accounting this, we can have an accurate countdown for the entire swap/bridge.

### DIFF
--- a/src/execution/StatusManager.ts
+++ b/src/execution/StatusManager.ts
@@ -170,19 +170,23 @@ export class StatusManager {
     switch (status) {
       case 'CANCELLED':
         currentProcess.doneAt = Date.now()
+        step.execution.pauseTimestamp = null
         break
       case 'FAILED':
         currentProcess.doneAt = Date.now()
         step.execution.status = 'FAILED'
+        step.execution.pauseTimestamp = null
         break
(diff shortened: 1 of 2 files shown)


src/execution/StatusManager.ts
```diff
@@ -179,6 +179,10 @@ export class StatusManager {
         currentProcess.doneAt = Date.now()
         break
       case 'PENDING':
+        if (step.execution.status === 'ACTION_REQUIRED') {
+          // the start of eoa confirmation process can be marked by the startAt of process
```
chybisov, line 183: Can we remove/rephrase eoa term and I guess `startAt` is an old name.

chybisov (changes requested):

PR #123, opened 2023-03-07
feat: allow user tracking by passing userId in header

Abhikumar98:
# Feature 

Allow user and integrator tracking by passing headers to every backend call

# Changes

In the SDK initialisation, now `userId` and `integrator` can be passed which would be later sent to backend for every request being made via the SDK.


src/types/internal.types.ts
```diff
@@ -35,6 +35,8 @@ export type Config = {
   defaultExecutionSettings: InternalExecutionSettings
   defaultRouteOptions: RouteOptions
   disableVersionCheck?: boolean
+  userId?: string
+  integrator: string
```
Addminus, line 39: why are we passing this to the ExecutionSettings?

Addminus (approved):

chybisov (approved):

PR #124, opened 2023-03-08
refactor: move logic to route execution manager + add node sdk sample

Abhikumar98:


src/Lifi.ts
```diff
@@ -327,10 +335,8 @@ export default class LIFI {
 
     const executionPromise = this.executeSteps(signer, clonedRoute, settings)
 
-    this.activeRouteDictionary[clonedRoute.id] = {
-      ...this.activeRouteDictionary[clonedRoute.id],
-      executionPromise,
-    }
+    this.activeRouteDictionary[clonedRoute.id].executionPromise =
+      executionPromise
```
chybisov, line 339: This code will still fail because we never assign an object to `this.activeRouteDictionary[clonedRoute.id]` - it's always `undefined`.

chybisov (approved):

PR #126, opened 2023-03-16
fix: skip tracking headers in non-lifi api calls

Abhikumar98:
## Issue

WIth the recent change, the tracking headers were being sent with every api call irrespective of LI.FI's APIs.

## Fix

Add an additional argument specifying whether the tracking header is required for that API call


src/helpers.ts
```diff
@@ -150,31 +154,33 @@ export const request = async <T = Response>(
       ...(options ?? {}),
     }
 
-    if (userId) {
-      updatedOptions.headers = {
-        ...options?.headers,
-        'X-LIFI-UserId': userId,
+    if (!skipTrackingHeaders) {
```
Abhikumar98, line 157: Adding this because tracking headers were going to npm registry as well and causing the request to fail.

src/helpers.ts
```diff
@@ -79,7 +79,10 @@ export const checkPackageUpdates = async (
   try {
     const pkgName = packageName ?? name
     const response = await request<{ version: string }>(
-      `https://registry.npmjs.org/${pkgName}/latest`
+      `https://registry.npmjs.org/${pkgName}/latest`,
+      undefined,
+      1,
+      true
```
chybisov, line 85: One more downside of the arguments nature of `request`. I also meant this when I suggested converting `request` arguments to an extended options object to avoid having `undefined` for skipped args. 🙂

chybisov (changes requested):

src/helpers.ts
Abhikumar98: True, this doesn't look good overall as well.

I've made a simple change in the structure which would accommodate all requirements while keeping `undefined` away but the types are a bit assertive.

```
interface ExtendedRequestInit extends RequestInit {
  retries?: number
  skipTrackingHeaders?: boolean
}

export const request = async <T = Response>(
  url: RequestInfo | URL,
  options = {
    retries: (this as unknown as ExtendedRequestInit)?.retries ?? 1,
    skipTrackingHeaders:
      (this as unknown as ExtendedRequestInit)?.skipTrackingHeaders ?? false,
  } as ExtendedRequestInit
): Promise<T> => {
```

Abhikumar98:
F.M.L.

TIL we can give typed default args this way 😭 
Thanks a lot @chybisov

src/helpers.ts
Abhikumar98: # F.M.L.

TIL we can give typed default args this way 😭 
Thanks a lot @chybisov

chybisov (approved):

PR #127, opened 2023-03-16
fix: return contextual transaction error message

Abhikumar98:
## ISSUE

No meaningful error messages are shown for transaction failure from smart contract.
eg. when token allowance is exhausted

## FIX

Using tenderly to decode the transaction errors via transaction hash and returning in the respective step


src/utils/parseError.ts
```diff
@@ -158,12 +162,38 @@ export const parseError = async (
...
-      )
+      try {
+        if (!step?.action.fromChainId) {
+          throw new Error('Signer is not defined.')
+        }
+
+        const response = await fetchTxErrorDetails(
+          e.transactionHash,
+          step?.action.fromChainId
+        )
```
chybisov, line 173: This method is async only because of ChainService, it's not really so async 😄 We should not use tenderly here and we can determine errors without it.

src/utils/parseError.ts
```diff
@@ -55,6 +58,7 @@ export const getTransactionNotSentMessage = async (
   step?: Step,
   process?: Process
 ): Promise<string> => {
+  console.trace('here')
```
chybisov, line 61: Please remove before commit 🙂

src/utils/parseError.ts
chybisov, line 173: This method is async only because of ChainService, it's not really so async 😄 We should not use tenderly here and we can determine errors without it.

chybisov (changes requested):

src/utils/parseError.ts
Abhikumar98: yes surely

chybisov: Sooo? 😅

src/utils/parseError.ts
```diff
@@ -158,12 +163,21 @@ export const parseError = async (
 
   switch (e.code) {
     case 'CALL_EXCEPTION':
+      if (e.reason === ERC20AllowanceError) {
```
chybisov, line 166: Can we use something like the above `e.reason?.includes('...')` here instead of creating a const just for this case? Also, we can consider refactoring those places above. We need to choose one approach for consistency 🙂

chybisov (changes requested):

src/utils/parseError.ts
chybisov: @Abhikumar98 Sorry, I'm afraid you missed the point here. I was referring to using [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) as a function there, not a property to compare with. 🙂

chybisov: Similar to this https://github.com/lifinance/sdk/pull/127/files#diff-4233cef87e84b84c5873476b8d547421036fa6283d27cb907461b9bdbd1a6af1R120-R121

src/utils/errors.ts
```diff
@@ -24,8 +24,11 @@ export enum LifiErrorCode {
   SlippageError = 1011,
   TransactionRejected = 1012,
   BalanceError = 1013,
+  AllowanceRequired = 1014,
 }
 
+export const ERC20AllowanceError = 'ERC20: transfer amount exceeds allowance'
+
```
chybisov, line 31: Please remove this and make it inline like the rest of the error strings.

src/utils/parseError.ts
Abhikumar98: ahh damn, I thought I'd removed them 😅

src/utils/errors.ts
Abhikumar98: I thought of keeping all raw strings in consts rather than using directly.

1. Reduces typos
2. Increases readability.

Any particular reason for using inline?

src/utils/errors.ts
```diff
@@ -24,6 +24,28 @@ export enum LifiErrorCode {
   SlippageError = 1011,
   TransactionRejected = 1012,
   BalanceError = 1013,
+  AllowanceRequired = 1014,
+}
+
+export enum GenericErrorType {
```
chybisov, line 30: I guess these errors come from ethers lib, so maybe `EthersErrorType`?

chybisov (changes requested):

chybisov (approved):

PR #128, opened 2023-03-17
feat: add callbacks to support custom transaction config

Abhikumar98:
# FEATURE

Add callbacks to support custom transaction config [LF-763]

## TL;DR

Developers can now customize the gas configuration of their on-chain transactions using a callback function. This allows for modification of gas limit, gas price, or priority fee before executing the transaction.

## DESCRIPTION

This feature enables developers to customize the gas configuration of their on-chain transactions using a callback function. When a transaction request is sent, the callback function returns the TransactionRequest object to the developer, who can modify the gas configuration such as max gas limit, gas price, or priority fee. Once the developer returns the modified TransactionRequest object, the transaction is executed with the updated configuration. 

This feature is useful for simulating transactions or setting custom gas limits and prices. It allows developers to fine-tune transaction requests based on their requirements.

## STEPS TO TEST

To test this feature, follow the steps below:

1. Add a callback function to the transaction request
2. Modify the TransactionRequest object to customize the gas configuration
3. Return the modified TransactionRequest object via the callback function
4. Execute the transaction with the updated configuration

## EXPECTED RESULTS

The transaction should be executed with the customized gas configuration provided by the developer.

## CHECKLIST

- [x] Performed a self-review of my code
- [x] Added comments to the code for better readability
- [x] Made sure that existing tests are still passing.

- [ ] Added tests that cover the new functionality

[LF-763]: https://lifi.atlassian.net/browse/LF-763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/allowance/checkAllowance.ts
+++ b/src/allowance/checkAllowance.ts
@@ -15,7 +15,9 @@ export const checkAllowance = async (
   settings: InternalExecutionSettings,
   chain: Chain,
   allowUserInteraction = false,
-  configCallback?: (tx: TransactionRequest) => Promise<TransactionRequest>
+  updateTransactionRequest?: (
+    tx: TransactionRequest
+  ) => Promise<TransactionRequest>
 ): Promise<void> => {
   // Ask the user to set an allowance
 
@@ -46,8 +48,8 @@ export const checkAllowance = async (
         to: step.estimate.approvalAddress,
       }
 
-      if (configCallback) {
-        const config = await configCallback(approvalRequest)
+      if (updateTransactionRequest) {
+        const config = await updateTransactionRequest(approvalRequest)
 
         approvalRequest.gasLimit = config.gasLimit
         approvalRequest.gasPrice = config.gasPrice
--- a/src/allowance/checkAllowance.ts
+++ b/src/allowance/checkAllowance.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-import { TransactionRequest } from '@ethersproject/abstract-provider/src.ts/index'
+import { TransactionRequest } from '@ethersproject/abstract-provider'
 import BigNumber from 'bignumber.js'
 import { constants, ContractTransaction, Signer } from 'ethers'
 import { getApproved, setApproval } from '../allowance/utils'
--- a/src/allowance/checkAllowance.ts
+++ b/src/allowance/checkAllowance.ts
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
+import { TransactionRequest } from '@ethersproject/abstract-provider/src.ts/index'
 import BigNumber from 'bignumber.js'
 import { constants, ContractTransaction, Signer } from 'ethers'
 import { getApproved, setApproval } from '../allowance/utils'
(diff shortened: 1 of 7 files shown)


src/types/internal.types.ts
```diff
@@ -24,6 +25,9 @@ export interface ExecutionParams {
   step: Step
   statusManager: StatusManager
   settings: InternalExecutionSettings
+  customConfigCallback?: (
```
chybisov, line 28: Please, let's find a better self-descriptive name for it. This one doesn't tell us what it is for.

chybisov:
Please link a task for it or describe why we make these changes.

src/allowance/checkAllowance.ts
```diff
@@ -13,7 +14,8 @@ export const checkAllowance = async (
   statusManager: StatusManager,
   settings: InternalExecutionSettings,
   chain: Chain,
-  allowUserInteraction = false
+  allowUserInteraction = false,
+  configCallback?: (tx: TransactionRequest) => Promise<TransactionRequest>
```
Addminus, line 18: maybe approvalConfigCallback is a more descriptive name?

Addminus (changes requested):

src/allowance/checkAllowance.ts
chybisov: I guess this is not only for approval but also for contract calls. https://lifi.atlassian.net/browse/LF-763

Maybe `updateTransactionRequest`?

Abhikumar98: updated the variable name

src/types/internal.types.ts
```diff
@@ -24,6 +25,9 @@ export interface ExecutionParams {
   step: Step
   statusManager: StatusManager
   settings: InternalExecutionSettings
+  updateTransactionRequest?: (
+    transactionRequest: TransactionRequest
+  ) => Promise<TransactionRequest>
```
chybisov, line 30: Does it make sense to have it as part of the `InternalExecutionSettings` like we have other hooks?

chybisov (changes requested):

src/types/internal.types.ts
Abhikumar98: Yes this already seems to be used at multiple places. I'm using this in my change.

src/allowance/utils.ts
```diff
@@ -9,13 +10,19 @@ import { fetchDataUsingMulticall, MultiCallData } from '../utils/multicall'
...
   const erc20 = new Contract(tokenAddress, ERC20_ABI, signer) as ERC20Contract
 
   try {
-    const approved = await erc20.allowance(signerAddress, contractAddress)
+    const approved = await erc20.allowance(signerAddress, contractAddress, {
+      gasLimit: transactionRequest?.gasLimit,
+      gasPrice: transactionRequest?.gasPrice,
+      maxFeePerGas: transactionRequest?.maxFeePerGas,
+      maxPriorityFeePerGas: transactionRequest?.maxPriorityFeePerGas,
+    })
```
chybisov, line 25: Can we pass the whole object here like 

`const approved = await erc20.allowance(signerAddress, contractAddress, transactionRequest)`

chybisov (changes requested):

chybisov:
Thanks for adding PR template 🙂

src/allowance/utils.ts
Abhikumar98: I only want to pick these gas related changes that the developer modifies.
Hence avoiding the usage of whole object.

There are variables like: `data` and `customData` which I assumed shouldn't be modified. Although the contract would already have these fields filled with the ongoing interaction.

chybisov (dismissed):

chybisov:
@Abhikumar98 please add tests 🙂

chybisov (approved):

PR #131, opened 2023-03-29
fix: errors in getStatus caused by the change of lifi/types

myz1237:
# Which Jira task belongs to this PR?

None

# What was broken / Why was this feature added?

Our discord bot will invoke getStatus function but encountered parameters error(missing fromChain, toChain or Bridge). Current version of lifi/types has covered new `GetStatusResponse` changes. I need to follow these changes to modify sdk and related tests

# How was it fixed / How was it implemented?

Delete outdated params checking and replace related tests with new ones

# How is the new behaviour / functionality tested?

Unit test

# Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] I have commented the code I wrote to make sure the next developers can understand it
- [x] I have added tests that cover the new functionality
- [x] I have made sure that existing tests are still passing
- [x] If this changed the API, I have updated the documentation (Gitbook and OpenAPI)

### DIFF
--- a/src/services/ApiService.ts
+++ b/src/services/ApiService.ts
@@ -239,36 +239,19 @@ const getContractCallQuote = async (
 }
 
 const getStatus = async (
-  { bridge, fromChain, toChain, txHash }: GetStatusRequest,
+  requestConfig: GetStatusRequest,
   options?: RequestOptions
 ): Promise<StatusResponse> => {
-  if (fromChain !== toChain && !bridge) {
-    throw new ValidationError(
-      'Parameter "bridge" is required for cross chain transfers.'
-    )
-  }
-
-  if (!fromChain) {
-    throw new ValidationError('Required parameter "fromChain" is missing.')
-  }
-
-  if (!toChain) {
-    throw new ValidationError('Required parameter "toChain" is missing.')
-  }
-
-  if (!txHash) {
+  if (!requestConfig.txHash) {
     throw new ValidationError('Required parameter "txHash" is missing.')
   }
 
   const config = ConfigService.getInstance().getConfig()
   try {
     const response = await request<StatusResponse>(
-      `${config.apiUrl}/status?${new URLSearchParams({
-        bridge,
-        fromChain,
-        toChain,
-        txHash,
-      } as Record<string, string>)}`,
+      `${config.apiUrl}/status?${new URLSearchParams(
+        requestConfig as unknown as Record<string, string>
+      )}`,
       {
         signal: options?.signal,
       }
--- a/src/services/ApiService.unit.spec.ts
+++ b/src/services/ApiService.unit.spec.ts
@@ -397,7 +397,7 @@ describe('ApiService', () => {
     })
   })
 
-  describe('getStatus', () => {
+  describe.only('getStatus', () => {
     const fromChain = ChainId.DAI
     const toChain = ChainId.POL
     const txHash = 'some tx hash'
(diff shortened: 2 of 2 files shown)


src/services/ApiService.ts
```diff
@@ -239,36 +239,19 @@ const getContractCallQuote = async (
...
   const config = ConfigService.getInstance().getConfig()
   try {
     const response = await request<StatusResponse>(
-      `${config.apiUrl}/status?${new URLSearchParams({
-        bridge,
-        fromChain,
-        toChain,
-        txHash,
-      } as Record<string, string>)}`,
+      `${config.apiUrl}/status?${new URLSearchParams(
```
Abhikumar98, line 252: let's move the formation of search params to a variable and then use that variable here.

Increase readability imo.

src/services/ApiService.unit.spec.ts
```diff
@@ -475,7 +440,28 @@ describe('ApiService', () => {
         it('call the server once', async () => {
           await ApiService.getStatus({ bridge, fromChain, toChain, txHash })
 
-          expect(mockedFetch).toHaveBeenCalledTimes(1)
+          await ApiService.getStatus({
+            bridge: undefined as unknown as string,
```
Abhikumar98, line 444: Is the type asserted here just for the test or it's expected in real-case scenario as well.

if `undefined` is an expected value, then the type for this should also be changed in lifi/types to:

```
bridge?: string;
fromChain?: ChainId;
toChain?: ChainId;
```

to avoid assertions.

src/services/ApiService.ts
myz1237: but it is not clean, I think
```ts
const getStatus = async (
  { txHash, fromChain, toChain, bridge }: GetStatusRequest,
  options?: RequestOptions
): Promise<StatusResponse> => {
  if (!txHash) {
    throw new ValidationError('Required parameter "txHash" is missing.')
  }
  const requestConfig: GetStatusRequest = { txHash }
  if (fromChain) {
    requestConfig.fromChain = fromChain
  }
  if (toChain) {
    requestConfig.toChain = toChain
  }
  if (bridge) {
    requestConfig.bridge = bridge
  }
  const config = ConfigService.getInstance().getConfig()
  try {
    const response = await request<StatusResponse>(
      `${config.apiUrl}/status?${new URLSearchParams(
        requestConfig as unknown as Record<string, string>
      )}`,
      {
        signal: options?.signal,
      }
    )

    return response
  } catch (e) {
    throw await parseBackendError(e)
  }
}
```

src/services/ApiService.unit.spec.ts
myz1237: real case.
lifi/types has been changed before. 
```ts
export interface GetStatusRequest {
    txHash: string;
    bridge?: string;
    fromChain?: number | string;
    toChain?: number | string;
}
```

src/services/ApiService.ts
Abhikumar98: Got it.
Agreed on the removal of multiple checks for the keys.

I was referring more towards the `URLSearchParams`.

```
try {

    // added below
    const queryParams = new URLSearchParams(requestConfig as unknown as Record<string, string>);

    const response = await request<StatusResponse>(
      `${config.apiUrl}/status?${queryParams}`,
      {
        signal: options?.signal,
      }
    )

    return response
  } catch (e) {
    throw await parseBackendError(e)
  }
```

Abhikumar98 (approved):

PR #134, opened 2023-04-24
fix: add tenderly for readable error messages

Abhikumar98:
# FEATURE

Adding Tenderly to get details Error messages #128

## TL;DR
Add Tenderly to get detailed error messages for all errors occurring during blockchain interaction.

## DESCRIPTION
NA

## STEPS TO TEST
N/A

## EXPECTED RESULTS
N/A

## CHECKLIST
- [x] I have performed a self-review of my code
- [x] I have commented the code I wrote to make sure the next developers can understand it
- [ ] I have added tests that cover the new functionality
- [x] I have made sure that existing tests are still passing


src/utils/parseError.ts
```diff
@@ -162,20 +163,41 @@ export const parseError = async (
...
+          return new TransactionError(
+            LifiErrorCode.AllowanceRequired,
+            e.reason,
+            errorMessage,
+            e.stack
+          )
+        }
+
+        // Error messages other than allowance error will be handled in catch block
+        throw new Error(e)
```
chybisov, line 192: Can we return the below `ProviderError` here instead of throwing a new one and then catching it?

chybisov (changes requested):

src/utils/parseError.ts
Abhikumar98: In the catch block, we are already throwing `ProviderError`. Hence just propagating regular error here.

chybisov: I mean why do we need `try...catch` here in the first place? 🙂

Abhikumar98: As per our previous discussions, to cover for any instance where tenderly API call fails.

If Tenderly API call fails, then resort to our default error message.

chybisov: Ah, okay, that makes sense.

src/utils/parseError.ts
```diff
@@ -162,20 +163,41 @@ export const parseError = async (
...
+        const errorMessage = response?.error_message ?? defaultErrorMessage
+
+        if (errorMessage?.includes(EthersErrorMessage.ERC20Allowance)) {
+          return new TransactionError(
+            LifiErrorCode.AllowanceRequired,
+            e.reason,
+            errorMessage,
+            e.stack
+          )
+        }
```
chybisov, line 189: I need one more clarification on this. 😄 Previously, we were checking if `e.reason` includes `EthersErrorMessage.ERC20Allowance` and now it looks like if tenderly returns an empty error message, then we will be checking `EthersErrorMessage.ERC20Allowance` against `defaultErrorMessage` returned by `getTransactionNotSentMessage` which will always be `false`. Am I missing something here?

chybisov (changes requested):

src/utils/parseError.ts
Abhikumar98: Resolved the issue. Thanks a lot for the careful review!

chybisov (approved):

PR #137, opened 2023-05-05
fix: increment gas on retry if out of gas

Abhikumar98:
# FEATURE

Retry with increased gas if transaction fails with out of gas error [LF-2862]

## TL;DR
Automatically retry transactions with increased gas if they fail with out of gas errors.

## DESCRIPTION
This feature automatically retries transactions with increased gas if they fail with out of gas errors. When attempting to send a transaction for bridging or swapping, if the initial gas is too low and the transaction fails with an out of gas error, the feature will automatically retry the transaction with increased gas to ensure that it is successfully processed.

## STEPS TO TEST
1. Reduce the gas to 21000 from Metamask before sending a transaction for bridging or swapping.
2. Observe that the transaction fails with an out of gas error.
3. The feature will automatically retry the transaction with increased gas.
4. Verify that the transaction is successfully processed.

## EXPECTED RESULTS
The feature should automatically retry transactions with increased gas if they fail with out of gas errors, and the transactions should be successfully processed.

## CHECKLIST
- [ ] I have performed a self-review of my code.
- [ ] I have commented the code I wrote to make sure the next developers can understand it.
- [ ] I have added tests that cover the new functionality.
- [ ] I have made sure that existing tests are still passing.

https://user-images.githubusercontent.com/35174288/236535777-e4a67fc2-ad13-4d41-9547-3fcd46431f22.mp4

[LF-2862]: https://lifi.atlassian.net/browse/LF-2862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/preRestart.ts
```diff
@@ -1,28 +1,38 @@
...
   )
   const isGasPriceError = route.steps[index].execution?.process.some(
     (p) => p.error?.code === LifiErrorCode.TransactionUnderpriced
   )
 
+  const { transactionRequest } = route.steps[index]
+
   if (isGasLimitError) {
+    if (transactionRequest) {
+      const gasLimit = signer.estimateGas(transactionRequest)
```
chybisov, line 30: Can this potentially throw something?

Abhikumar98: This ideally shouldn't throw any errors until there's an issue from RPC or an incorrect `transactionRequest` being sent to it.

But to be on safer side, wrapped in `try...catch` block and also handling the promises correctly now.

src/LiFi.unit.spec.ts
```diff
@@ -1,28 +1,86 @@
...
+          gasAmount: '123',
+          gasAmountUSD: '123',
+          gasPrice: '123',
+          gasToken: '123',
+          gasUsed: '123',
+        },
+      })
+    ),
+  },
+}))
```
chybisov, line 39: It looks okay, but can we please use `msw` here, like in the rest of the tests, to mock responses from our BE? The same applies to all tests, if any.

src/LiFi.unit.spec.ts
```diff
@@ -1,28 +1,86 @@
...
+
+vi.mock('./allowance/utils', () => ({
+  getApproved: vi.fn(() => Promise.resolve([])),
+}))
+
+// vi.mock('./execution/utils', () => ({
+// waitForReceivingTransaction: vi.fn(() => Promise.resolve()),
+// getProcessMessage: vi.fn(() => 'message'),
+// checkStepSlippageThreshold: vi.fn(() => true),
+// }))
```
chybisov, line 60: Please remove unnecessary comments when you finish with the PR.

test/fixtures.ts
```diff
@@ -168,3 +168,97 @@ export const buildRouteObject = ({
     state: 'NOT_INSURABLE',
   },
 })
+
+export const mockChainsResponse = [
```
chybisov, line 172: This should act as `msw` response.

chybisov (changes requested):

src/LiFi.unit.spec.ts
Abhikumar98: Sure, I'll make this change

test/fixtures.ts
Abhikumar98: alright, making this change!

src/LiFi.unit.spec.ts
```diff
@@ -211,4 +269,49 @@ describe('LIFI SDK', () => {
...
+    it('should pick up gas estimation from signer', async () => {
+      const route = buildRouteObject({
+        step,
+      })
+
+      await lifi.executeRoute(signer, route)
+
+      expect(signer.sendTransaction).toHaveBeenCalledWith({
+        gasLimit: '125000',
+        // TODO: Check the cause for gasLimit being outside transactionRequest. Currently working as expected in widget
```
Addminus, line 283: Do you want to fix that in this PR or in another one?

If you want to leave that as TODO here, please mention it in the PR description

Addminus (changes requested):

src/utils/preRestart.ts
```diff
@@ -1,28 +1,41 @@
...
 
+  const { transactionRequest } = route.steps[index]
+
   if (isGasLimitError) {
+    if (transactionRequest) {
+      let gasLimit = transactionRequest.gasLimit
+
+      try {
+        gasLimit = await signer.estimateGas(transactionRequest)
+      } catch (error) {}
```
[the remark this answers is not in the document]
Abhikumar98: I recently had a conversation with @chybisov where he mentioned that this is best being ignored.

although, at that time I'd mistakenly written `console.log` instead of `console.error`.

I'd still prefer having an error being logged. because the rest of the error are being handled in custom error handlers.

wdyt?
cc: @chybisov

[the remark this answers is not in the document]
chybisov: We should have a fallback in case it throws, but we don't log anything in the SDK now, so I see no reason to log this one for now, but maybe we should come up with something in the future. We should especially avoid `console.error` as it can annoy some envs.

test/fixtures.ts
chybisov: @Abhikumar98 please don't resolve conversations until fixed 😄

Abhikumar98: Now the mockResponse is being used in the msw mocks. Does this need to be relocated to another file?

src/LiFi.unit.spec.ts
Abhikumar98: Duplicate. Resolving this

chybisov (changes requested):

src/LiFi.unit.handlers.ts
```diff
@@ -0,0 +1,38 @@
...
+      )
+  ),
+  rest.get(`${config.apiUrl}/chains`, async (_, response, context) =>
+    response(
+      context.status(200),
+      context.json({
+        chains: mockChainsResponse,
+      })
+    )
+  ),
```
chybisov, line 34: Couldn't find where we use this endpoint in tests. 😅

chybisov (changes requested):

yarn.lock
```diff
@@ -6602,6 +6603,15 @@ __metadata:
...
 
+"tsc@npm:^2.0.4":
+  version: 2.0.4
+  resolution: "tsc@npm:2.0.4"
+  bin:
+    tsc: bin/tsc
+  checksum: 50b10240887424c66454687f0fca4e319713b35cb44f0dab228719eb4dd757251f83c3d79e60d29dfbe90c2d6f1272f793b83e75a9fca2c622b4213ad3eb8cb1
+  languageName: node
+  linkType: hard
+
```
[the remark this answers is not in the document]
Abhikumar98: I was testing with tsc. I'll remove this.
Got pushed by mistake

src/LiFi.unit.handlers.ts
Abhikumar98: This is being called from inside lifi.execute

src/LiFi.unit.spec.ts
```diff
@@ -210,4 +259,28 @@ describe('LIFI SDK', () => {
...
+          data: '0xdata',
+          from: '0x552008c0f6870c2f77e5cC1d2eb9bdff03e30Ea0',
+          gasLimit: '682701',
+          gasPrice: '0x27c01c1727',
+          to: '0x1231DEB6f5749EF6cE6943a275A1D3E7486F4EaE',
+          value: '0x0600830dbc7f5bf7',
+        },
+      })
+    })
+  })
```
chybisov, line 285: @Abhikumar98 I just realized that we test functionality for `RouteExecutionManager` here and LIFI class just extends it. Does it make sense to move these tests to RouteExecutionManager.unit.spec.ts?

src/LiFi.unit.handlers.ts
chybisov: > The api endpoints for this is being called within the lifi.execute method. Hence mocking the response for that over here.

I can't seem to find that place at the first glance. Please share 🙏😄

chybisov (changes requested):

src/LiFi.unit.spec.ts
Abhikumar98: Yes we can do that. That'd would be better

src/LiFi.unit.handlers.ts
Abhikumar98: This is being called while creating new instance of `StepExecutionManager`.

Inside the execute method, we are using `ChainService.getInstance()`, which fetches all the chains from the Backend.

src/utils/preRestart.ts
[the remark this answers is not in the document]
Addminus: alright! We should make sure that a failure here does not break the process. We should see that estimation as an addition. If we don't set anything, metamask sets its own estimations.
So it's sort of a nice to have if it works but not breaking and falling back to metamask defaults if it does not work. It looks like the code does exactly that.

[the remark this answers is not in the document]
Addminus: jup true

src/execution/StepExecutionManager.ts
```diff
@@ -153,6 +153,14 @@ export class StepExecutionManager {
...
               customConfig.maxPriorityFeePerGas
             transactionRequest.maxFeePerGas = customConfig.maxFeePerGas
+          } else {
+            const estimatedGasLimit = await signer.estimateGas(
+              transactionRequest
+            )
+
+            transactionRequest.gasLimit = `${Math.round(
+              Number(estimatedGasLimit) * 1.25
+            )}`
```
chybisov, line 163: Reviewing this once again, I realized that the return type from `estimateGas` is `BigNumber`, but we convert it to a plain `Number` here, and sure, it's an edge case for gas estimation, but this potentially can lead to an error if we get a big number that exceeds `Number.MAX_SAFE_INTEGER`. I think we should switch to `BigInt` here.

Also, when we switch to `viem`, this function starts to return `BigInt`, which just reinforces my words 🙂

chybisov (changes requested):

src/LiFi.unit.handlers.ts
chybisov: @Abhikumar98 do we need that huge mock here, or can we just mock something simple just to avoid a call to BE?

src/execution/StepExecutionManager.ts
chybisov: `${BigInt(estimatedGasLimit) * 125n / 100n}`

src/LiFi.unit.handlers.ts
Abhikumar98: We'll need Polygon, because that's where the swap is happening. I'll remove the other one

src/execution/StepExecutionManager.ts
Abhikumar98: Hey @chybisov,
Pushed the changes for this :)

_`/prerestart.ts`_
CleanShot 2023-05-17 at 16 47 20

_`StepExecutionManager.ts`_
CleanShot 2023-05-17 at 16 46 50

src/utils/preRestart.ts
```diff
@@ -1,28 +1,45 @@
...
+
+      if (gasLimit) {
+        transactionRequest.gasLimit = `${
+          (BigInt(gasLimit?.toString()) * 125n) / 100n
```
chybisov, line 38: Unnecessary optional chaining here due to the above check. Also, passing undefined to BigInt will throw TypeError.

src/execution/StepExecutionManager.ts
```diff
@@ -153,6 +153,16 @@ export class StepExecutionManager {
...
               customConfig.maxPriorityFeePerGas
             transactionRequest.maxFeePerGas = customConfig.maxFeePerGas
+          } else {
+            const estimatedGasLimit = await signer.estimateGas(
+              transactionRequest
+            )
+
+            if (estimatedGasLimit) {
+              transactionRequest.gasLimit = `${
+                (BigInt(estimatedGasLimit?.toString()) * 125n) / 100n
```
chybisov, line 163: Same unnecessary optional chaining.

chybisov (changes requested):

chybisov (approved):

PR #143, opened 2023-05-18
fix: estimate gas price from signer LF-2417

Abhikumar98:
# FIX

Handle Transaction underpriced error [LF-2417]

## TL;DR
Automatically handle "Transaction Underpriced" errors by retrying the transaction with an increased gas price.

## DESCRIPTION
This feature handles the scenario where a transaction fails with a "Transaction Underpriced" error. Upon encountering such an error, the feature will automatically retry the transaction with an increased gas price. The latest gas price will be fetched from the signer, and it will be increased by 25% before sending the transaction forward.

## STEPS TO TEST
1. Initiate a swap or bridge transaction.
2. While at the Metamask screen, reduce the Max Base Fee and Priority Fee
3. Confirm the transaction and observe the "Transaction Underpriced" error.
4. Upon the next retry the transaction will be retried with an increased gas price.
5. Verify that the retried transaction is sent with a gas price estimated from the signer and increased by 25%.

## EXPECTED RESULTS
Upon encountering a "Transaction Underpriced" error, the feature should automatically retry the transaction with an increased gas price based on the latest signer estimate.

## CHECKLIST
- [x] I have performed a self-review of my code.
- [x] I have commented the code I wrote to make it easily understandable.
- [x] I have added tests that cover the new functionality.
- [x] I have ensured that existing tests are passing.

[LF-2417]: https://lifi.atlassian.net/browse/LF-2417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/execution/StepExecutionManager.ts
+++ b/src/execution/StepExecutionManager.ts
@@ -154,22 +154,24 @@ export class StepExecutionManager {
               customConfig.maxPriorityFeePerGas
             transactionRequest.maxFeePerGas = customConfig.maxFeePerGas
           } else {
-            const estimatedGasLimit = await signer.estimateGas(
-              transactionRequest
-            )
+            try {
+              const estimatedGasLimit = await signer.estimateGas(
+                transactionRequest
+              )
 
-            if (estimatedGasLimit) {
-              transactionRequest.gasLimit = `${
-                (BigInt(estimatedGasLimit.toString()) * 125n) / 100n
-              }`
-            }
+              if (estimatedGasLimit) {
+                transactionRequest.gasLimit = `${
+                  (BigInt(estimatedGasLimit.toString()) * 125n) / 100n
+                }`
+              }
 
-            // Fetch latest gasPrice from provider and use it
-            const gasPrice = await signer.getGasPrice()
+              // Fetch latest gasPrice from provider and use it
+              const gasPrice = await signer.getGasPrice()
 
-            if (gasPrice) {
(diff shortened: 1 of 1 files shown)


Abhikumar98:
Avoiding the default increase in gas price in StepExecutionManager.ts as it'd always result in an increase gas price.

src/execution/StepExecutionManager.ts
```diff
@@ -163,6 +163,13 @@ export class StepExecutionManager {
                 (BigInt(estimatedGasLimit.toString()) * 125n) / 100n
               }`
             }
+
+            // Fetch latest gasPrice from provider and use it
+            const gasPrice = await signer.getGasPrice()
```
chybisov, line 168: If this fails the whole swap via the widget will throw an unknown error. We should wrap all calls to `signer.getGasPrice` and `signer.estimateGas` try...catch and fallback to BE values/previous logic.

chybisov (changes requested):

src/execution/StepExecutionManager.ts
chybisov: I got an error while testing 😄

Abhikumar98: Got it. I've pushed the changes

Addminus (approved):
lets go

PR #148, opened 2023-06-08
fix: add custom gas limit for approvals LF-2862

Abhikumar98:
Add custom gas limit for token approvals
missed implementation for approvals in #137

### DIFF
--- a/src/allowance/utils.ts
+++ b/src/allowance/utils.ts
@@ -1,7 +1,7 @@
 import { TransactionRequest } from '@ethersproject/abstract-provider'
 import { ChainId, Token } from '@lifi/types'
 import BigNumber from 'bignumber.js'
-import { Contract, ContractTransaction, Signer } from 'ethers'
+import { Contract, ContractTransaction, Signer, ethers } from 'ethers'
 import ChainsService from '../services/ChainsService'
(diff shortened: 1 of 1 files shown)


src/allowance/utils.ts
```diff
@@ -29,15 +29,30 @@ export const getApproved = async (
...
+  )
+
+  try {
+    const estimatedGasLimit = await signer.estimateGas(transactionRequest)
+    if (estimatedGasLimit) {
+      const formattedGasLimit = ethers.utils.parseEther(
+        `${(BigInt(estimatedGasLimit.toString()) * 125n) / 100n}`
+      )
+
+      transactionRequest.gasLimit = formattedGasLimit
```
chybisov, line 51: Are we sure about bumping the gas limit by 25% right away?

chybisov: In other parts, we bump it if there is an error with current estimation.

chybisov (changes requested):

src/allowance/utils.ts
Abhikumar98: This change was done based on an internal discussion. Sharing the discord link in DM.

chybisov (approved):

chybisov (approved):

PR #150, opened 2023-06-22
feat: multisig wallets support

Abhikumar98:
Safe integration change for SDK. More details later


src/execution/StepExecutionManager.ts
```diff
@@ -30,6 +34,13 @@ export class StepExecutionManager {
     statusManager,
     settings,
   }: ExecutionParams): Promise<Execution> => {
+    const isSafeSigner = !!(signer.provider as any)?.provider?.safe?.safeAddress
+
+    const safeProvider = (signer.provider as any)?.provider?.safe
+    const safeProviderSDK = (signer.provider as any)?.provider?.sdk
+
+    console.log({ safeProvider, safeProviderSDK })
```
Addminus, line 42: we should remove logs

src/execution/StepExecutionManager.ts
```diff
@@ -62,6 +75,84 @@ export class StepExecutionManager {
     let process = statusManager.findOrCreateProcess(step, currentProcessType)
 
     if (process.status !== 'DONE') {
+      // TODO: update @lifi/types repo for these changes [DO NOT MERGE IF YOU SEE THIS]
+
+      console.log('NOT DONE YET!!')
+
+      console.log({ isSafeSigner, process, step })
+
+      if (isSafeSigner && step.execution.status === 'SAFE_SIGNING_PENDING') {
```
Addminus, line 84: Can we rename the status message to "MULTISIGNATURE_SIGNING_PENDING"?
that way its more general

Addminus (changes requested):

src/execution/StepExecutionManager.ts
chybisov: I don't like the idea of a safe wallet spread across all our repos. Is there a way to have it only within Jumper? What if we need another multisig wallet support tomorrow? We should stay wallet agnostic in the SDK.

src/execution/StepExecutionManager.ts
```diff
@@ -34,12 +32,8 @@ export class StepExecutionManager {
...
     settings,
   }: ExecutionParams): Promise<Execution> => {
-    const isSafeSigner = !!(signer.provider as any)?.provider?.safe?.safeAddress
-
-    const safeProvider = (signer.provider as any)?.provider?.safe
-    const safeProviderSDK = (signer.provider as any)?.provider?.sdk
-
-    console.log({ safeProvider, safeProviderSDK })
+    const config = ConfigService.getInstance().getConfig()
+    const isSafeSigner = !!config.multiSigConfig?.isMultiSigSigner
```
chybisov, line 36: Should be `isMultisigSigner`

src/execution/StepExecutionManager.ts
```diff
@@ -76,35 +70,38 @@ export class StepExecutionManager {
...
-      if (isSafeSigner && step.execution.status === 'SAFE_SIGNING_PENDING') {
-        console.log('Alrighty coming here')
-
+      if (
+        isSafeSigner &&
+        step.execution.status === 'MULTISIG_SIGNING_PENDING'
+      ) {
         const multiSigProcess = step.execution.process.find(
-          (p) => p.status === 'SAFE_SIGNING_PENDING'
+          (p) => p.status === 'MULTISIG_SIGNING_PENDING'
```
chybisov, line 78: We agreed not to have a separate status for execution/process and reuse ACTION_REQUIRED one 👀

src/execution/StepExecutionManager.ts
```diff
@@ -76,35 +70,38 @@ export class StepExecutionManager {
...
+        ) {
+          throw new Error('MultiSig process is undefined')
+        }
+        if (!config.multiSigConfig?.getMultiSigTransactionDetails) {
+          throw new Error(
+            '"getMultiSigTransactionDetails" is missing in MultiSig config.'
+          )
         }
 
         const multiSigInternalTxHash = multiSigProcess.multiSigInternalTxHash
```
chybisov, line 93: Let's rename this to `multisigTxHash`

src/execution/StepExecutionManager.ts
```diff
@@ -114,14 +111,12 @@ export class StepExecutionManager {
...
-          console.log('transaction is rejected')
+        if (response.status === 'FAILED') {
           process = statusManager.updateProcess(step, process.type, 'FAILED', {
             error: {
-              message: 'Transaction failed',
-              htmlMessage: 'Transaction failed',
-              code: LifiErrorCode.TransactionFailed,
+              message: response.message,
+              htmlMessage: response.message,
+              code: LifiErrorCode.MultiSigTransactionFailed,
```
chybisov, line 119: Why do we need a custom error here?

chybisov (changes requested):

package.json
```diff
@@ -77,6 +77,7 @@
     "@ethersproject/abi": "^5.7.0",
     "@ethersproject/contracts": "^5.7.0",
     "@lifi/types": "^6.0.0",
+    "@safe-global/safe-apps-sdk": "^7.11.0",
```
chybisov, line 80: Should be removed from here as per discussion 🙂

src/execution/StepExecutionManager.ts
```diff
@@ -62,6 +70,66 @@ export class StepExecutionManager {
...
+
+        if (response.status === 'SUCCESS') {
+          statusManager.updateExecution(step, 'DONE')
+          process = statusManager.updateProcess(step, process.type, 'DONE', {
+            txHash: response.txHash,
+            txLink:
+              fromChain.metamask.blockExplorerUrls[0] + 'tx/' + response.txHash,
+          })
+
+          return step.execution!
```
chybisov, line 115: Are you sure this can be `DONE` on this stage? 👀

chybisov (changes requested):

src/execution/StepExecutionManager.ts
Abhikumar98: Changed back to PENDING

src/execution/StepExecutionManager.ts
Abhikumar98: I thought we did PENDING :/

package.json
Abhikumar98: Removing

src/execution/StepExecutionManager.ts
Abhikumar98: yes, this is when the status from SAFE Interaction is completed

src/execution/StepExecutionManager.ts
Abhikumar98: the variable or the property?
or both?

src/execution/StepExecutionManager.ts
chybisov: > I thought we did PENDING :/

Yes, PENDING/ACTION_REQUIRED, but without an introduction of a new one.

src/execution/StepExecutionManager.ts
chybisov: > yes, this is when the status from SAFE Interaction is completed

I think not because in that case we say that the execution of the whole step is DONE, which is not true at this stage.

I assume we have SUCCESS status for multisig transaction, which means all required signers have signed the transaction and we now should wait for it to be mined. Am I missing something here?

src/execution/StepExecutionManager.ts
Abhikumar98: Imo, having separate context for error is better to avoid introducing logic in presentational components or functions returning plain text, such as `getProcessMessage` function.

To communicate with precise error message and details, went ahead with custom error message.

src/execution/StepExecutionManager.ts
Abhikumar98: Kinda yes, afaik the SAFE status would be success only when the transaction in mined and all the other execution process is done.

src/execution/StepExecutionManager.ts
```diff
@@ -62,7 +70,72 @@ export class StepExecutionManager {
...
 
     if (process.status !== 'DONE') {
+      const multisigProcess = step.execution.process.find(
+        (p) => !!p.multisigTxHash
+      )
+
       try {
+        if (isMultisigSigner && multisigProcess) {
+          if (!multisigProcess) {
+            throw new Error('MultiSig process is undefined')
```
chybisov, line 80: MultiSig -> Multisig

src/execution/StepExecutionManager.ts
```diff
@@ -62,7 +70,72 @@ export class StepExecutionManager {
...
+              }
+            )
+          }
+
+          if (response.status === 'FAILED') {
+            throw new TransactionError(
+              LifiErrorCode.TransactionFailed,
+              '',
+              '',
+              ''
```
chybisov, line 125: Why is this empty? 🙂

src/execution/StepExecutionManager.ts
```diff
@@ -178,17 +251,42 @@ export class StepExecutionManager {
...
           // STEP 4: Wait for the transaction
-          process = statusManager.updateProcess(step, process.type, 'PENDING', {
-            txHash: transaction.hash,
-            txLink:
-              fromChain.metamask.blockExplorerUrls[0] +
-              'tx/' +
-              transaction.hash,
-          })
+          if (isMultisigSigner) {
+            statusManager.updateExecution(step, 'ACTION_REQUIRED')
```
chybisov, line 255: We need to double-check if we change execution to ACTION_REQUIRED somewhere because maybe it's not needed.

src/types/internal.types.ts
```diff
@@ -42,6 +42,18 @@ export type Config = {
   userId?: string
   integrator: string
   widgetVersion?: string
+  multisigConfig?: MultiSigConfig
+}
+
+export interface MultiSigTxDetails {
```
chybisov, line 48: MultiSig -> Multisig

src/types/internal.types.ts
```diff
@@ -42,6 +42,18 @@ export type Config = {
...
+  status: 'SUCCESS' | 'FAILED' | 'PENDING' | 'CANCELLED'
+  message: string
+  txHash?: string
+}
+
+export interface MultiSigConfig {
```
chybisov, line 54: MultiSig -> Multisig

src/execution/StepExecutionManager.ts
```diff
@@ -62,7 +70,72 @@ export class StepExecutionManager {
...
+      )
+
       try {
+        if (isMultisigSigner && multisigProcess) {
+          if (!multisigProcess) {
+            throw new Error('MultiSig process is undefined')
+          }
+          if (!config.multisigConfig?.getMultisigTransactionDetails) {
+            throw new Error(
+              '"getMultiSigTransactionDetails()" is missing in MultiSig config.'
```
chybisov, line 84: MultiSig -> Multisig

src/utils/errors.ts
```diff
@@ -26,6 +26,7 @@ export enum LifiErrorCode {
   BalanceError = 1013,
   AllowanceRequired = 1014,
   InsufficientFunds = 1015,
+  MultiSigTransactionFailed = 1016,
```
chybisov, line 29: MultiSig -> Multisig

chybisov (changes requested):

src/execution/StepExecutionManager.ts
Abhikumar98: umm, I didn't exactly know the error to fill in, I'll write the same message itself. Lmk if it works

src/types/internal.types.ts
```diff
@@ -42,18 +42,32 @@ export type Config = {
...
 
-export interface MultiSigTxDetails {
+export interface MultisigTxDetails {
   status: 'SUCCESS' | 'FAILED' | 'PENDING' | 'CANCELLED'
   message: string
   txHash?: string
 }
 
-export interface MultiSigConfig {
+export interface SendTransactionsMultisigResponse {
```
chybisov, line 54: `MultisigTransactionResponse`?

src/execution/StepExecutionManager.ts
```diff
@@ -43,27 +54,108 @@ export class StepExecutionManager {
...
+            throw new TransactionError(
+              LifiErrorCode.TransactionFailed,
+              'Multisig transaction failed.'
+            )
+          }
+
+          if (response.status === 'CANCELLED') {
+            throw new TransactionError(
+              LifiErrorCode.TransactionRejected,
+              'Transaction was rejected by users'
```
chybisov, line 153: Let's be consistent with messages and dots at the end.

src/execution/StepExecutionManager.ts
```diff
@@ -175,19 +267,76 @@ export class StepExecutionManager {
...
+
+          if (
+            shouldBatchTransactions &&
+            config.multisigConfig?.sendBatchTransaction
+          ) {
+            const { to, data, value } = await signer.populateTransaction(
+              transactionRequest
+            )
+
+            const isValidTransaction = value !== undefined && to && data
```
chybisov, line 279: Are we sure about an undefined check like this? Maybe we can check `!value`?

src/execution/StepExecutionManager.ts
Abhikumar98: alright!
Removing the `.` from everywhere

src/execution/StepExecutionManager.ts
```diff
@@ -43,27 +54,109 @@ export class StepExecutionManager {
...
+        shouldBatchTransactions
       )
+
+      if (populatedTransaction) {
+        const { to, data } = populatedTransaction
+
+        if (to && data) {
+          // allowance doesn't need value
+          const cleanedPopulatedTransaction: BaseTransaction = {
+            value: '0x00',
```
chybisov, line 82: Avoid, please if possible.

src/execution/StepExecutionManager.ts
chybisov: Please remove as per discussion

src/execution/StepExecutionManager.ts
Abhikumar98: replaced this with empty string

src/execution/StepExecutionManager.ts
chybisov: Maybe return BigNumber(0)?

src/execution/StepExecutionManager.ts
```diff
@@ -43,27 +62,79 @@ export class StepExecutionManager {
...
+          const multisigTxHash = multisigProcess.multisigTxHash
+
+          if (!multisigTxHash) {
+            // need to check what happens in failed tx
+            throw new ValidationError(
+              'Multisig internal transaction hash is undefined.'
+            )
+          }
+
+          await updateSafeRouteProcess(
```
Abhikumar98, line 128: Abstracted the logic of fetching status from SAFE and processing response to be used when the transaction is initiated as well.

In the `fetchMultisigStatus` function the recursion is present, hence we need to trigger this to initiate the polling.


And currently the polling logic is present only upon resuming the route (here).

Abstracting the logic for this and reusing it post initiation: https://github.com/lifinance/sdk/pull/150/files#diff-bc44d69ee7c901547808f63d12468d9523a358d036d8f69db430969460cec5e1R321

src/allowance/utils.ts
```diff
@@ -52,7 +64,7 @@ export const setApproval = async (
     }
   } catch (error) {}
 
-  return signer.sendTransaction(transactionRequest)
+  return erc20.approve(contractAddress, amount)
```
Abhikumar98, line 67: revert to `signer.sendTransaction(transactionRequest)`

chybisov (approved):

chybisov (approved):

PR #155, opened 2023-07-19
fix: parsing params in available connections api request

Abhikumar98:
# FIX

Parsing all params to get the available connections [LF-3826]

## TL;DR

Fix the issue where only selected params with array value were being parsed, leaving out other array-based params. Now all array params and all values are parsed and passed in the API request.

## DESCRIPTION

Previously, there was an issue where only selected parameters with array values were being parsed, resulting in the exclusion of other array-based parameters. This fix addresses that issue by ensuring that all array parameters and their respective values are parsed and correctly passed in the API request.

## TESTING

No testing is required as this fix is already integrated and tested.

## CHECKLIST

- [x] I have performed a self-review of my code.
- [ ] I have commented the code I wrote to ensure clarity.
- [x] I have made sure that existing tests related to this fix are passing.
- [x] I have verified that the fix resolves the issue described in the ticket.

[LF-3826]: https://lifi.atlassian.net/browse/LF-3826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/services/ApiService.unit.spec.ts
+++ b/src/services/ApiService.unit.spec.ts
@@ -704,16 +704,16 @@ describe('ApiService', () => {
         fromToken: findDefaultToken(CoinKey.USDC, ChainId.BSC).address,
         toToken: findDefaultToken(CoinKey.USDC, ChainId.OPT).address,
         allowBridges: ['connext', 'uniswap', 'polygon'],
-        allowExchanges: ['a', 'b', 'c'],
-        denyBridges: ['testValue1', 'testValue2'],
-        preferBridges: ['testValue1', 'testValue2'],
-        denyExchanges: ['testValue1', 'testValue2'],
-        preferExchanges: ['testValue1', 'testValue2'],
+        allowExchanges: ['1inch', 'ParaSwap', 'SushiSwap'],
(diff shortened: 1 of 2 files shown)


src/services/ApiService.unit.spec.ts
```diff
@@ -704,11 +704,16 @@ describe('ApiService', () => {
         fromToken: findDefaultToken(CoinKey.USDC, ChainId.BSC).address,
         toToken: findDefaultToken(CoinKey.USDC, ChainId.OPT).address,
         allowBridges: ['connext', 'uniswap', 'polygon'],
+        allowExchanges: ['a', 'b', 'c'],
+        denyBridges: ['testValue1', 'testValue2'],
+        preferBridges: ['testValue1', 'testValue2'],
+        denyExchanges: ['testValue1', 'testValue2'],
+        preferExchanges: ['testValue1', 'testValue2'],
```
chybisov, line 711: Should we add real data to this test?

chybisov (changes requested):

chybisov (approved):

PR #162, opened 2023-08-17
fix: remove updateTransactionRequestHook from allowance read call

Abhikumar98:
https://lifi.atlassian.net/browse/LF-4515


src/allowance/checkAllowance.unit.spec.ts
Addminus, line 1: Why did you delete the unit test?

Abhikumar98: updated and brought back the regular allowance call

Addminus (approved):

chybisov (approved):
@Abhikumar98 please use meaningful messages as PR titles because these titles then become CHANGELOG records 🙂

Also, https://www.conventionalcommits.org/

@Addminus I noticed this issue while migrating to viem, which is already fixed in the viem PR. I thought it didn't break anything in the existing release, but it appears it does 🙂

Also, we need to rewrite the test. It doesn't do much now, but probably leave this to after the viem era 🙂

PR #164, opened 2023-09-13
feat: add multi-contract calls API endpoint

Abhikumar98:
# ⚠️ Deprecation

The method: `getContractCallQuote` is being deprecated and `getContractCallsQuote` must be used instead.

## Added Multi-contract call api endpoint

new method `getContractCallsQuote` is added to enable user to get quotes from multiple contract calls

### DIFF
--- a/src/services/ApiService.ts
+++ b/src/services/ApiService.ts
@@ -2,6 +2,7 @@ import {
   ConnectionsRequest,
   ConnectionsResponse,
   ContractCallQuoteRequest,
+  ContractCallsQuoteRequest,
   GasRecommendationRequest,
   GasRecommendationResponse,
   GetStatusRequest,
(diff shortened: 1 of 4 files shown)


Abhikumar98:
The `@lifi/types` package has been bumped to latest version `8.7.0`, to pull in the new types

chybisov (approved):

src/services/ApiService.ts
```diff
@@ -222,6 +223,68 @@ const getContractCallQuote = async (
...
+    'toAmount',
+    'contractCalls',
+  ]
+
+  if (requestConfig.contractCalls.length === 0) {
+    throw new ValidationError(`Parameter "contractCalls" is empty.`)
+  }
+
+  requiredParameters.forEach((requiredParameter) => {
+    if (!requestConfig[requiredParameter]) {
```
0xsebn, line 248: wdyt about using "in" operator instead of the if check ? 

`if (requiredParameter in requestConfig)`

0xsebn (approved):

PR #167, opened 2023-10-03
fix: wallet analytics API request changes

Abhikumar98:
There are changes from backend in the request config for the analytics API, 
https://github.com/lifinance/lifi-backend/pull/2770

[API Reference](https://apidocs.li.fi/reference/get_analytics-wallets-wallet-address-1)

### DIFF
--- a/src/services/ApiService.ts
+++ b/src/services/ApiService.ts
@@ -28,6 +28,7 @@ import {
   Token,
   ToolsRequest,
   ToolsResponse,
+  WalletAnalyticsRequest,
 } from '../types'
 import { ValidationError } from '../utils/errors'
 import { parseBackendError } from '../utils/parseError'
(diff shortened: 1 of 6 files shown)


src/types/internal.types.ts
```diff
@@ -191,3 +191,10 @@ export interface Signature {
   r: string
   s: string
 }
+
+export interface WalletAnalyticsRequest {
+  wallet_address: string
+  fromTimestamp: number
+  toTimestamp: number
+  integrator?: string
+}
```
chybisov, line 200: Why do we end up with a type here again? I remember we moved it to the types package.
 Also, why do we start using snake_case?

chybisov (changes requested):

src/types/internal.types.ts
Abhikumar98: This is just the request type for the endpoint, but just checked that all the request types as in /types package.

moving it there.

src/services/ApiService.ts
```diff
@@ -522,14 +523,38 @@ const getAvailableConnections = async (
...
   )
 
+  url.searchParams.append(
+    'toTimestamp',
+    walletAnalyticsRequest.toTimestamp.toString()
+  )
+
+  if (config.integrator) {
+    url.searchParams.append('integrator', config.integrator)
+  }
```
chybisov, line 554: We add this automatically inside request function 🙂

chybisov (changes requested):

src/services/ApiService.ts
Abhikumar98: Inside the request method, we are passing it in headers and here the backend as a field in query params.

Seeing that, I additionally added the integrator as a query param here. If it's the same and backend always picks up from the headers then I'll go ahead and remove this.

Abhikumar98: Upon discussion, resorting to passing integrator via headers in request function and migrating to query param in v3 probably

chybisov (approved):

0xsebn (approved):

PR #182, opened 2024-03-20
fix: default imports for parcel bundler

Addminus:
Ticket: https://lifi.atlassian.net/browse/LF-7519

Description
- importing balance logic by calling `import * as balance from './balance` caused issues in parcel production builds

Changes:
- changing import of balance handling logic due to issues with using the SDK in the parcel bundler
- updating tests with same import changes and updating the mocking logic for the balance functions

I created a small js app with parcel and react in order to reproduce this issue locally. The changes in this PR fix the issues. 
I also used the local build  in a vite react js project after making the changes just to be sure.

### DIFF
--- a/src/LiFi.ts
+++ b/src/LiFi.ts
@@ -38,7 +38,7 @@ import {
   getTokenApproval,
   revokeTokenApproval,
 } from './allowance'
-import * as balance from './balance'
+import balance from './balance'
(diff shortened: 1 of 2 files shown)


DNR500 (approved):
Thanks for the fix

mathiasmoeller (approved):
LGTM 👌

PR #191, opened 2024-06-11
feat: improve handling of errors

DNR500:
Jira: [LF-8182](https://lifi.atlassian.net/browse/LF-8182)

[LF-8182]: https://lifi.atlassian.net/browse/LF-8182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/helpers.ts
```diff
@@ -62,7 +62,8 @@ export const convertQuoteToRoute = (step: LiFiStep): Route => {
 
 export const fetchTxErrorDetails = async (txHash: string, chainId: number) => {
   const response = await request<TenderlyResponse>(
-    `https://api.tenderly.co/api/v1/public-contract/${chainId}/tx/${txHash}`
+    `https://api.tenderly.co/api/v1/public-contract/${chainId}/tx/${txHash}`,
+    { disableLiFiErrorCodes: true }
```
DNR500, line 66: We use the request function in a couple of places to make requests from sources outside of the LiFi domain - This flag opts out of using the LiFi error codes in situations like this.

src/services/api.ts
```diff
@@ -48,20 +47,15 @@ export const getToken = async (
...
-        chain,
-        token,
-      } as Record<string, string>)}`,
-      {
-        signal: options?.signal,
-      }
-    )
-    return response
-  } catch (e) {
-    throw await parseBackendError(e)
```
DNR500, line 63: parseBackendError as basically do two main things - using response.json() to access the response body message and then also creating an error with sets of error codes and error types dependant on status code of the response.

Both these tasks have been moved into the HTTPError class - with the async method to process the response.json() being triggered in the request function. The try/catch shouldn't be needed here anymore.

src/core/EVM/EVMStepExecutor.ts
```diff
@@ -480,7 +484,7 @@ export class EVMStepExecutor extends BaseStepExecutor {
       })
       this.statusManager.updateExecution(step, 'FAILED')
       console.warn(e)
```
chybisov, line 486: Let's remove this warn

src/core/EVM/EVMStepExecutor.ts
```diff
@@ -87,9 +86,13 @@ export class EVMStepExecutor extends BaseStepExecutor {
...
-      throw new TransactionError(
-        LiFiErrorCode.WalletChangedDuringExecution,
-        errorMessage
+      throw await parseEVMStepErrors(
+        new TransactionError(
+          LiFiErrorCode.WalletChangedDuringExecution,
+          errorMessage
+        ),
+        step,
+        process
```
chybisov, line 95: Do you think it makes sense to parse it here, as this code is usually wrapped inside another try-catch, and we will end up parsing it twice? Also, I think we should only parse errors before throwing them to the top as a last step.

src/core/EVM/parseEVMStepErrors.ts
chybisov, line 1: We discussed naming them as `parseEVMError.ts` and `parseSVMError.ts`/`parseSolanaError.ts`

src/helper.unit.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+describe('helpers', () => {
+  describe('checkPackageUpdates', () => {
+    beforeEach(() => {
+      vi.spyOn(global, 'fetch').mockResolvedValue({
+        json: () => Promise.resolve({ version: latestVersion }),
+      } as Response)
+
+      vi.spyOn(console, 'warn').mockImplementation(() => {})
+    })
```
chybisov, line 22: Should we move this to msw as well? 😄

src/utils/errors/index.ts
chybisov, line 1: How about we move errors outside of utils to have it's own folder?

chybisov (changes requested):

src/utils/errors/index.ts
DNR500: I like that idea

src/core/EVM/parseEVMStepErrors.ts
DNR500: Makes sense

src/core/EVM/EVMStepExecutor.ts
DNR500: Yes its the only console.warn I think. Makes sense to remove it

src/helper.unit.spec.ts
DNR500: Can do

DNR500: Thinking about it the latest version endpoint isn't under our control and isn't so likely to change - I think mocking with fetch is fine

src/core/EVM/EVMStepExecutor.ts
DNR500: The checkWallet function isn't always used in a part of the executeStep method that isn't cover by parseEVMStep Errors in the try catch there. I think if we look to simplify the executeStep method future we could look to reduce the use of parseEVMStepErrors then.

PR #222, opened 2024-12-06
Execute send transaction from Viem client

carlosfebres:
## Summary
To enable execution from more complex wallet clients (e.g. smart wallet account clients), the send transaction Viem function is usually overwritten to handle sending these transactions. Viem currently doesn't support the execution of smart accounts in their default send transaction function.

Viem uses a `getAction` util function to execute a given function on the client or default to Viem's default function, which I believe should be used within the SDK.

### DIFF
--- a/src/core/EVM/EVMStepExecutor.ts
+++ b/src/core/EVM/EVMStepExecutor.ts
@@ -5,6 +5,7 @@ import type {
 } from '@lifi/types'
 import type { Client, Hash, SendTransactionParameters } from 'viem'
 import { getAddresses, sendTransaction } from 'viem/actions'
+import { getAction } from 'viem/utils'
 import { config } from '../../config.js'
 import { LiFiErrorCode } from '../../errors/constants.js'
 import { TransactionError, ValidationError } from '../../errors/errors.js'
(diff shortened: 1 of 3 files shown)


chybisov:
@carlosfebres hey, thank you for the PR. Could you please describe your use case more and which smart wallets you use? 

PR makes sense, we need to see if there are any more places we need to use getAction 👀

carlosfebres:
Hey @chybisov! We are using a [Kernel wallet](https://github.com/zerodevapp/kernel) to hold the funds, and we want it to use LiFi without using an EOA as a middleware. The changes required to make it work are minimal; this will be very useful for anyone trying LiFi with a smart wallet account.

chybisov:
@carlosfebres hey, thank you for getting back and making the PR. I replaced some other wallet actions with `getAction` util.

caffeinum:
this sounds amazing! is there a guide/integration example?

for our smart account wallet, currently we're pretending to be a multisig, and batch transactions manually

I would love to migrate to a cleaner approach.

caffeinum:
@chybisov sure! We're already using LiFi SDK in our smart wallet app, however currently I have a clunky wrapper-adapter that pretends to be a multisig, to support batching. If smart wallet usecase is the one already supported by SDK, I would love to re-integrate properly.

We also have a group chat in telegram with Li.Fi team – is it best to ask there?

PR #234, opened 2025-02-04
fix: fixed evm address compare check

Defus-E:


Defus-E:
> @Defus-E hey, why do we need this?

Encountered error `The wallet address that requested the quote does not match the wallet address attempting to sign the transaction`. As EVM addresses are not case-sensitive, I think it worth to compare lowercased versions of addresses. 

For example, viem client class stores address version like `0x7C9881b88af63F6d8c74732727b2e1B05801fA14`, but when I request a lifi's route I pass `"fromAddress": "0x7c9881b88af63f6d8c74732727b2e1b05801fa14"` - lowercased variant of the same address. Despite this, i receive an error that addresses are different

chybisov:
@Defus-E oh yeah, good catch, thank you!

chybisov (approved):

chybisov:
@Defus-E please upgrade to [v3.5.3](https://github.com/lifinance/sdk/releases/tag/v3.5.3)

PR #245, opened 2025-04-02
Avoid message signing with flag

carlosfebres:
## Summary
Recently, token approvals were done by signing a message for tokens supporting [EIP-2612](https://eips.ethereum.org/EIPS/eip-2612). This is great for avoiding executing a transaction to approve tokens only, but it doesn't work for all wallets. 

Permit doesn't work for smart wallet accounts since EIP-2612 omits ERC-1271, which is often used to "sign" messages by smart wallet accounts. 

To still allow smart wallet accounts to use LiFi's SDK, I propose a configuration flag `disableMessageSigning` that skips the Permit logic and always executes a token approval.

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -114,6 +114,7 @@ export interface ExecutionOptions {
   updateRouteHook?: UpdateRouteHook
   updateTransactionRequestHook?: TransactionRequestUpdateHook
   executeInBackground?: boolean
+  disableMessageSigning?: boolean
   /**
    * @deprecated
    */
--- a/src/core/EVM/EVMStepExecutor.ts
+++ b/src/core/EVM/EVMStepExecutor.ts
@@ -237,7 +237,7 @@ export class EVMStepExecutor extends BaseStepExecutor {
     // Check if step requires permit signature and will be used with relayer service
     const isRelayerTransaction = isRelayerStep(step)
 
-    // Check if wallet can sign messages, usefully for smart wallet accounts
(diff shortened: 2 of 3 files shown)


src/core/types.ts
```diff
@@ -114,6 +114,7 @@ export interface ExecutionOptions {
   updateRouteHook?: UpdateRouteHook
   updateTransactionRequestHook?: TransactionRequestUpdateHook
   executeInBackground?: boolean
+  disableMessageSigning?: boolean
```
chybisov, line 117: Since this is EVM specific for now, should we move this flag to be EVM executor only (EVMStepExecutor)?

chybisov (commented):
@carlosfebres hey, thank you for contribution! 

We are skipping message signing when the wallet supports transaction batching through EIP-5792, is it something you can make use of? If it's not the case, then the PR makes total sense.

src/core/types.ts
carlosfebres: I thought about this; I think it looks cleaner having it in the `ExecutionOptions`; otherwise, we'd need to have this flag in the EVM's configuration and pass it down to the EVMStepExecutor.

carlosfebres:
> We are skipping message signing when the wallet supports transaction batching through EIP-5792, is it something you can make use of? If it's not the case, then the PR makes total sense.

We can't because that would require all RPC providers we use among multiple chains to support EIP-5792.

PR #250, opened 2025-04-22
fix: add new error codes

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-13094

### DIFF
--- a/src/errors/constants.ts
+++ b/src/errors/constants.ts
@@ -9,7 +9,6 @@ export enum ErrorName {
   UnknownError = 'UnknownError',
   SlippageError = 'SlippageError',
   HTTPError = 'HTTPError',
-  ThirdPartyError = 'ThirdPartyError',
 }
 
 export type ErrorCode = LiFiErrorCode
--- a/src/errors/constants.ts
+++ b/src/errors/constants.ts
@@ -9,6 +9,8 @@ export enum ErrorName {
   UnknownError = 'UnknownError',
   SlippageError = 'SlippageError',
   HTTPError = 'HTTPError',
+  RateLimitExceeded = 'RateLimitExceeded',
+  ThirdPartyError = 'ThirdPartyError',
 }
 
 export type ErrorCode = LiFiErrorCode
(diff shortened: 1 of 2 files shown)


src/errors/constants.ts
```diff
@@ -37,6 +39,9 @@ export enum LiFiErrorCode {
   TransactionConflict = 1020,
   TransactionNotFound = 1021,
   TransactionRejected = 1022,
+  TooManyRequests = 1023,
```
chybisov, line 42: Let's name this one as `RateLimitExceeded`

src/errors/constants.ts
```diff
@@ -9,6 +9,8 @@ export enum ErrorName {
   UnknownError = 'UnknownError',
   SlippageError = 'SlippageError',
   HTTPError = 'HTTPError',
+  RateLimitExceeded = 'RateLimitExceeded',
+  ThirdPartyError = 'ThirdPartyError',
```
chybisov, line 13: Both of these will be under `ServerError`, so no need to add new names.

src/errors/constants.ts
```diff
@@ -37,6 +39,9 @@ export enum LiFiErrorCode {
   TransactionConflict = 1020,
   TransactionNotFound = 1021,
   TransactionRejected = 1022,
+  TooManyRequests = 1023,
+  UnprocessableEntity = 1024,
+  FailedDependency = 1025,
```
chybisov, line 44: Let's name this one as `ThirdPartyError`

src/errors/constants.ts
```diff
@@ -9,6 +9,7 @@ export enum ErrorName {
   UnknownError = 'UnknownError',
   SlippageError = 'SlippageError',
   HTTPError = 'HTTPError',
+  ThirdPartyError = 'ThirdPartyError',
```
chybisov, line 12: We don't use this error name, so let's remove it and proceed. 🚀

chybisov (approved):

PR #261, opened 2025-05-26
fix: add public client fallback for wallet client failures on EVM RPC calls

chybisov:
Ticket: [LF-13859](https://lifi.atlassian.net/browse/LF-13859)

### DIFF
--- a/src/core/EVM/getActionWithFallback.ts
+++ b/src/core/EVM/getActionWithFallback.ts
@@ -1,0 +1,58 @@
+import type {
+  Account,
+  Chain,
+  Client,
+  PublicActions,
+  RpcSchema,
+  Transport,
+  WalletActions,
+} from 'viem'
+import { getAction } from 'viem/utils'
+import { getPublicClient } from './publicClient.js'
+
+/**
+ * Executes an action with a fallback to public client if the wallet client fails due to rate limiting
+ * or similar recoverable errors.
+ *
+ * Note: Only falls back to public client if the initial client was a wallet client (has an account address).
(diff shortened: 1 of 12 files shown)


src/core/EVM/getActionWithFallback.ts
```diff
@@ -0,0 +1,58 @@
...
+  returnType,
+>(
+  walletClient: client,
+  actionFn: (_: client, parameters: parameters) => returnType,
+  name: keyof PublicActions | keyof WalletActions | (string & {}),
+  params: parameters
+): Promise<Awaited<returnType>> => {
+  try {
+    return await getAction(walletClient, actionFn, name)(params)
+  } catch (error: unknown) {
```
effie-ms, line 43: Some thoughts: It would be probably great to check whether it is "rate limiting or similar recoverable errors" by error code or anything else. While this might not be that critical, I am also thinking about cases when an action is not supported by the public client and can only be done by the wallet client. Then falling back to the public client would not really make sense. But then it is also a matter of usage of that wrapper (will need to remember what kind of fallback is out there). 

I think for now it is okay.

effie-ms (approved):

src/core/EVM/getActionWithFallback.ts
chybisov: @tomiiide we use a generic `Client` type everywhere, which might be both wallet and public client in theory, but in most cases where we deal with a wallet client we can trace it. Yeah, it can be confusing at first, I agree. Do you have any ideas?

PR #269, opened 2025-06-16
fix: add gas limit buffer for permit flow

chybisov:
[LF-14152](https://lifi.atlassian.net/browse/LF-14152)


tomiiide (approved):
Looks good!

PR #271, opened 2025-06-23
feat(solana): improve transaction confirmation logic with polling

tomiiide:
Closes https://lifi.atlassian.net/browse/LF-12980

This PR updates the logic of `confirmTransaction()` on Solana to poll an API instead of using a websocket connection.

### DIFF
--- a/src/core/Solana/sendAndConfirmTransaction.ts
+++ b/src/core/Solana/sendAndConfirmTransaction.ts
@@ -1,8 +1,6 @@
 import type {
-  Connection,
   SendOptions,
   SignatureResult,
-  TransactionSignature,
   VersionedTransaction,
 } from '@solana/web3.js'
 import bs58 from 'bs58'
@@ -26,7 +24,6 @@ export async function sendAndConfirmTransaction(
   const connections = await getSolanaConnections()
 
   const signedTxSerialized = signedTx.serialize()
-
   // Create transaction hash (signature)
   const txSignature = bs58.encode(signedTx.signatures[0])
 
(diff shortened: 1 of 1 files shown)


src/core/Solana/sendAndConfirmTransaction.ts
```diff
@@ -41,71 +42,86 @@ export async function sendAndConfirmTransaction(
...
-            abortSignal: abortController.signal,
-          },
-          'confirmed'
-        )
-        .then((result) => result.value)
-
-      let signatureResult: SignatureResult | null = null
-      let blockHeight = await connection.getBlockHeight('confirmed')
+  const pollPromises = connections.map(async (connection) => {
+    const timeout = 60000
```
chybisov, line 46: It would be nice to keep the previous logic and keep it based on the block hash expiration, instead of custom timeouts.

src/core/Solana/sendAndConfirmTransaction.ts
```diff
@@ -41,71 +42,86 @@ export async function sendAndConfirmTransaction(
...
-      return signatureResult
+async function pollTransactionConfirmation(
+  txtSig: TransactionSignature,
+  connection: Connection
+): Promise<TransactionSignature> {
+  // 15 second timeout
+  const timeout = 15000
+  // 5 second retry interval
+  const interval = 5000
+  const startTime = Date.now()
```
chybisov, line 91: Same here. It would be nice to keep 1-second timeout for this to speed up transaction landing.

src/core/Solana/sendAndConfirmTransaction.ts
```diff
@@ -41,71 +42,86 @@ export async function sendAndConfirmTransaction(
...
+  const checkStatus = async () => {
+    try {
+      const status = await connection.getSignatureStatuses([txtSig])
+      return status?.value[0]?.confirmationStatus === 'confirmed'
     } catch (error) {
-      if (abortController.signal.aborted) {
-        return Promise.reject(new Error('Confirmation aborted.'))
-      }
-      throw error
+      console.warn('Error checking transaction status:', error)
```
chybisov, line 98: We shouldn't have console warnings.

chybisov (changes requested):

chybisov (approved):

PR #281, opened 2025-07-08
feat: add UNS support

tomiiide:
This PR adds support for fetching address records from Unstoppable Domains for all chains.

closes https://lifi.atlassian.net/browse/LF-12980
closes https://lifi.atlassian.net/browse/LF-3239

https://github.com/user-attachments/assets/2cf17fe0-46ba-406e-92b5-45e4f92ff4cb


src/core/EVM/resolveEVMAddress.ts
```diff
@@ -0,0 +1,9 @@
+import { ChainId } from '@lifi/types'
+import { getENSAddress } from './getENSAddress.js'
+import { getUNSAddress } from './uns/getUNSAddress.js'
+
+export async function resolveEVMAddress(
+  name: string
+): Promise<string | undefined> {
+  return (await getENSAddress(name)) || (await getUNSAddress(name, ChainId.ETH))
```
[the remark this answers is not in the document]
tomiiide: I'm fetching the ETH address stored in that name, maybe I should use the `ChainType` instead.

src/core/Solana/resolveSolanaAddress.ts
```diff
@@ -0,0 +1,12 @@
+import { ChainType } from '@lifi/types'
+import { resolveUNSAddress } from '../EVM/uns/resolveUNSAddress.js'
```
chybisov, line 2: Oh, Solana ecosystem should not have any imports form EVM, it breaks dependency separation

src/core/uns/resolveUNSAddress.ts
```diff
@@ -0,0 +1,110 @@
...
+      abi: UNSProxyReaderABI,
+      address: proxyAddress,
+      functionName: 'exists',
+      args: [BigInt(name)],
+    } as const
+
+    const exists = await readContractAction(existsReadContractParameters)
+
+    if (!exists) {
+      return null
```
effie-ms, line 83: Looking at line 36: if we eventually resolve nulls to undefined, maybe we could do it immediately? like `return undefined` from `getUnsAddress` where we have nulls now, and then just `return address` in line 36?

src/core/uns/resolveUNSAddress.ts
```diff
@@ -0,0 +1,110 @@
...
+
+type GetUnsAddressReturnType = Address | null
+
+async function getUnsAddress(
+  client: Client,
+  params: GetUnsAddressParameters
+): Promise<GetUnsAddressReturnType> {
+  const { name, chain } = params
+
+  // TODO: For more robust resolution, we should construct the keys based on the token and not the chain
```
effie-ms, line 57: is it a blocker?

src/core/uns/resolveUNSAddress.ts
```diff
@@ -0,0 +1,110 @@
...
+  const keys = [`crypto.${chain}.address`]
+
+  try {
+    const chainId = client.chain?.id
+    if (!chainId) {
+      throw new Error('Chain ID not available')
```
effie-ms, line 63: if these errors are caught just to return null, should we maybe just return nulls then (since we will never see any messages eventually)?

src/core/uns/resolveUNSAddress.ts
```diff
@@ -0,0 +1,110 @@
...
+  } catch (_) {
+    // ignore
+    return
+  }
+}
+
+type GetUnsAddressParameters = {
+  chain: string
+  name: string
+  token?: string
```
effie-ms, line 46: is `token` used anywhere?

tomiiide: No, leftover from trying to do a token based resolution, would remove.

effie-ms (approved):

PR #288, opened 2025-07-28
feat: add getWalletBalances()

effie-ms:
Add endpoint to fetch cached wallet balances from the backend.

### DIFF
--- a/src/services/balance.ts
+++ b/src/services/balance.ts
@@ -1,10 +1,8 @@
 import type {
   GetWalletBalanceExtendedResponse,
-  GetWalletBalanceResponse,
   RequestOptions,
   Token,
   TokenAmount,
-  WalletToken,
   WalletTokenExtended,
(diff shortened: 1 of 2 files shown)


src/services/balance.ts
```diff
@@ -103,3 +104,27 @@ export const getTokenBalancesByChain = async (
...
+): Promise<{ [chainId: number]: TokenAmount[] }> => {
+  if (!walletAddress) {
+    throw new ValidationError('Missing walletAddress.')
+  }
+
+  const response = await request<{
+    balances: { [chainId: number]: TokenAmount[] }
+    walletAddress: string
+    limit: number
+  }>(`${config.get().apiUrl}/wallets/${walletAddress}/balances`, {
```
chybisov, line 126: There is a `GetWalletBalanceResponse` type that we can use here.

chybisov (approved):

PR #327, opened 2025-11-19
refactor(solana)!: upgrade from web3.js to solana kit

tomiiide:
## Which Jira task is linked to this PR?  
https://lifi.atlassian.net/browse/LF-13181

_Explain the reasoning behind the implementation. Were there alternative approaches? Why was this solution chosen?_  

The solana provider
1. drops the outdated `@solana/web.js` library for the new `@solana/kit`
2. replaces the wallet / adapter type from `@solana/wallet-adapter-base` to the new `@wallet-standard/wallet-standard`.

### Breaking changes
`getWalletAdapter` config option is renamed to `getWallet`

_If applicable, attach screenshots, GIFs, or videos to showcase the functionality, UI changes, or bug fixes._  


tomiiide:
> @tomiiide could you please rebase and make the PR only about Solana Kit changes 🙏 It would be hard to review 273 files again 😅

@chybisov done!

packages/sdk-provider-solana/src/actions/sendAndConfirmTransaction.ts
```diff
@@ -86,28 +105,40 @@ export async function sendAndConfirmTransaction(
           }
 
           await sleep(400)
+
+          if (!abortController.signal.aborted) {
+            blockHeight = await connection
+              .getBlockHeight({
+                commitment: 'confirmed',
+              })
+              .send()
```
chybisov: We didn't have a block height update here to reduce the number of requests to the RPC. I think it is fine to update it every second for both cases when we send a transaction below. Wdyt?

tomiiide: Hmm, I separated them to avoid race conditions as we are sharing `blockheight` between two loops.
The polling loop is updating the blockheight, while the sending loop is reading it.

Yes, we get 2x more rpc calls, but better code quality.

tomiiide: It does make sense to update once every second when resending, Since the worst case scenario is minimal during a race condition, it just feels dirty.

packages/sdk-provider-solana/package.json
```diff
@@ -48,10 +48,9 @@
   },
   "dependencies": {
     "@lifi/sdk": "workspace:*",
-    "@noble/curves": "^1.9.7",
-    "@solana/wallet-adapter-base": "^0.9.27",
-    "@solana/web3.js": "^1.98.4",
-    "bs58": "^6.0.0"
+    "@solana/compat": "^5.0.0",
+    "@solana/kit": "^5.0.0",
+    "@solana/web3.js": "^1.98.4"
```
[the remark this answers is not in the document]
chybisov: I think we should return the adapter like before from the widget, so no versioned txs conversion should happen there. Can we manage this in the SDK without depending on `@solana/web3.js`?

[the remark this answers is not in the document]
tomiiide: If we return the adapter like before then, the sdk would still need to convert txns to/from versioned txns as the adapters only handle versioned txn types.

We can switch the adapter types to wallet standard `uiWallet` type like it's done in the [`@solana/kit` example.](https://github.com/anza-xyz/kit/blob/main/examples/react-app/src/components/SolanaSignTransactionFeaturePanel.tsx) but that introduces the problem we had before where we have to manage `currentWallet` state as it's not done by `wallet-standard/react`.

There is an [official library](https://github.com/solana-foundation/framework-kit) that does this, and [the first version](https://www.npmjs.com/package/@solana/react-hooks) has been released, but we need this [PR](https://github.com/solana-foundation/framework-kit/pull/4) merged to be able to fully support it. 

Therefore my suggestion is to have the transaction compatibility layer live on `solana-widget-provider` for now, and after `framework-kit` is mature, we can completely migrate solana wallet management to `wallet-standard` without issues.

packages/sdk-provider-solana/src/utils/KeypairWallet.int.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+const base58Encoder = getBase58Encoder()
+
+const createTestClient = (wallet: KeypairWalletAdapter) => {
+  const client = createClient({
+    integrator: 'lifi-sdk-test',
+  })
+
+  client.setProviders([
+    SolanaProvider({
+      getWallet: async () => wallet,
```
effie-ms: Nice it's `getWallet()` now, we will just need to rename it on the widget side as well from `getWalletAdapter()`. I wonder if we could have `getWallet()` for all the ecosystems eventually. 
(EVM and Bitcoin providers have `getWalletClient()` now).
Not a part of this PR though :)

packages/sdk-provider-solana/src/utils/KeypairWalletAdapter.ts
```diff
@@ -1,100 +1,133 @@
...
+      if (
+        signatureDictionaries.length === 0 ||
+        !signatureDictionaries[0][this._signer.address]
+      ) {
+        throw new Error('Failed to sign transaction')
+      }
+
+      const signedTransaction = getBase64EncodedWireTransaction(transaction)
 
-  async signMessage(message: Uint8Array): Promise<Uint8Array> {
```
effie-ms: So we won't be exposing signing messages anymore, right?

tomiiide: Yeah, we won't.

tomiiide: It would a class that extends the wallet standard wallet, and would be called using the features standard.

packages/sdk-provider-solana/src/utils/KeypairWallet.int.spec.ts
chybisov: @tomiiide I'd keep getWallet(), as we have the same naming convention for Sui, which also uses the Wallet Standard approach.

PR #329, opened 2025-11-26
feat(v3-solana): add support for jito bundles

tomiiide:
## Which Jira task is linked to this PR?  
https://lifi.atlassian.net/browse/LF-16736

_Explain the reasoning behind the implementation. Were there alternative approaches? Why was this solution chosen?_  

This PR adds support for Jito bundles to enable more complex transactions on the Solana network

Key changes
- Adds JitoConnection class extending Solana's Connection with bundle-specific methods (simulateBundle, sendBundle, getTipAccounts)
- Adds automatic detection of rpcs with jito support by checking for `getTipAccounts` method availability.
- Adds `sendAndConfirmBundle` with parallel sending and polling support like the existing `sendAndConfirmTransaction`
- Updates `solanaStepExecutor` to handle bundles and single transaction depending on the sdk config and transactions

_If applicable, attach screenshots, GIFs, or videos to showcase the functionality, UI changes, or bug fixes._  

An example bundle -> https://explorer.jito.wtf/bundle/4b63086c8da6af20e5e715e9b5dc3278abb4d99442a6a64deea285f3b6cf1a34

### DIFF
--- a/src/errors/httpError.unit.spec.ts
+++ b/src/errors/httpError.unit.spec.ts
@@ -100,29 +100,26 @@ describe('HTTPError', () => {
           '[ValidationError] Request failed with status code 400 Bad Request',
       },
     ],
-  ])(
-    'should present correctly %s',
-    async (_, requestOptions, statusCode, statusText, expected) => {
-      const mockResponse = {
-        status: statusCode,
-        statusText,
-        json: expected.jsonFunc,
-      } as Response
+  ])('should present correctly %s', async (_, requestOptions, statusCode, statusText, expected) => {
+    const mockResponse = {
+      status: statusCode,
+      statusText,
+      json: expected.jsonFunc,
+    } as Response
 
-      const error = new HTTPError(mockResponse, url, requestOptions)
+    const error = new HTTPError(mockResponse, url, requestOptions)
 
-      expect(error.status).toEqual(statusCode)
-      expect(error.message).toEqual(expected.initialMessage)
-      expect(error.url).toEqual(url)
-      expect(error.fetchOptions).toEqual(requestOptions)
+    expect(error.status).toEqual(statusCode)
+    expect(error.message).toEqual(expected.initialMessage)
+    expect(error.url).toEqual(url)
+    expect(error.fetchOptions).toEqual(requestOptions)
 
-      expect(error.type).toEqual(expected.type)
-      expect(error.code).toEqual(expected.code)
+    expect(error.type).toEqual(expected.type)
+    expect(error.code).toEqual(expected.code)
 
-      await error.buildAdditionalDetails()
+    await error.buildAdditionalDetails()
 
-      expect(error.responseBody).toEqual(expected.responseBody)
-      expect(error.message).toEqual(expected.builtMessage)
-    }
-  )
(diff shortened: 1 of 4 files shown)


src/core/Solana/connection.ts
effie-ms: If we have no connections on some reason, we are throwing a null error here, just a note.
Not sure how plausible this situation is, but if this is the case, maybe better to check the size of `connections` or return null instead of throwing.

src/core/Solana/jito/JitoConnection.ts
```diff
@@ -0,0 +1,122 @@
...
+      },
+      body: JSON.stringify({
+        jsonrpc: '2.0',
+        id: 1,
+        method,
+        params,
+      }),
+    })
+
+    const data = await response.json()
```
effie-ms: should we check `response.ok` before that?

src/services/api.ts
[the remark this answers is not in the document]
tomiiide: @chybisov should we have this?

src/core/Solana/connection.ts
tomiiide: it is highly unlikely this happens, but handled regardless

src/core/Solana/jito/sendAndConfirmBundle.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/**
+ * Send and confirm a bundle of transactions using Jito
+ * Automatically selects a Jito-enabled RPC connection and polls for confirmation
+ * across multiple Jito RPCs in parallel
+ * @param signedTransactions - Array of signed transactions
+ * @returns Bundle ID, transaction signatures, and confirmation results
+ */
+export async function sendAndConfirmBundle(
```
chybisov: This implementation diverges from how we handle standard transactions. Can we align it with the existing pattern so it's easier to maintain and follow?

1.	We should avoid running simulations inside a loop.
2.	The same loop should be used for connection handling, retrieving block height, confirming transactions, etc., rather than splitting the logic.

src/core/Solana/connection.ts
```diff
@@ -1,8 +1,23 @@
...
+const detectCluster = (rpcUrl: string): Cluster => {
+  const url = rpcUrl.toLowerCase()
+  if (url.includes('devnet')) {
+    return 'devnet'
+  }
+  if (url.includes('testnet')) {
+    return 'testnet'
+  }
+  return 'mainnet-beta'
+}
```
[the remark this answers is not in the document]
tomiiide: Would remove

src/core/Solana/jito/JitoConnection.ts
```diff
@@ -0,0 +1,238 @@
...
+    return this.rpcRequest<BundleStatusResult>('getBundleStatuses', [bundleIds])
+  }
+
+  /**
+   * Confirm a bundle by polling its status until confirmed or timeout
+   * @param bundleId - The bundle UUID to confirm
+   * @param timeoutMs - Maximum time to wait in milliseconds (default: 60000)
+   * @returns true if bundle was confirmed, false if timeout
+   */
+  async confirmBundle(
```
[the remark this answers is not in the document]
tomiiide: We don't, removed.

PR #375, opened 2026-04-22
fix(sdk-provider-solana): balance aggregation, confirmation races, and RPC ordering

dhruvja:
## Summary
                                                                                                                                                                                                                                                                                              
  A handful of correctness bugs in `@lifi/sdk-provider-solana`:                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                              
  - **`getSolanaBalance`** — `tokenAmounts[mint] = amount` overwrote earlier entries, so wallets with multiple token accounts for the same mint (common for USDC) had balances reported as the last-seen account only. Now sums them. Also only reports `0n` for an SPL mint when **both** the
   Token and Token-2022 program queries succeeded, since a mint may live in whichever program's query failed (e.g. PYUSD on Token-2022).
  - **`sendAndConfirmTransaction` / `sendAndConfirmBundle`** — `Promise.any(promises).catch(() => null)` was being fed promises that could fulfil with `null` on a per-RPC timeout. Because `null` is a fulfilled value, `Promise.any` resolved to `null` as soon as the first RPC gave up,   
  discarding valid confirmations from slower RPCs. The inner map now throws a sentinel on `null`, so `Promise.any` skips it and waits for a real confirmation (or rejects once every RPC has given up).                                                                                       
  - **`SolanaJitoWaitForTransactionTask`** — empty `signedTransactions` now throws `TransactionUnprepared` before we call `sendAndConfirmBundle`, matching the guard already present in `SolanaStandardWaitForTransactionTask`.
  - **`rpc/registry.ts`** — `getSolanaRpcs` / `getJitoRpcs` returned `Array.from(map.values())`, which ignored the URL order from `client.getRpcUrlsByChainId`. `callWithRetry` iterates in array order, so retries were effectively random across runs. Both functions now derive the list from the URL array and filter out entries lost to LRU eviction.  

- [ ] This pull request is focused and addresses a single problem.  

### DIFF
--- a/packages/sdk-provider-solana/src/actions/getSolanaBalance.ts
+++ b/packages/sdk-provider-solana/src/actions/getSolanaBalance.ts
@@ -115,7 +115,8 @@ const getSolanaBalanceDefault = async (
         value.account.data.parsed.info
       const amount = BigInt(tokenAccount.tokenAmount.amount)
       if (amount > 0n) {
-        tokenAmounts[tokenAccount.mint] = amount
+        tokenAmounts[tokenAccount.mint] =
+          (tokenAmounts[tokenAccount.mint] ?? 0n) + amount
       }
       return tokenAmounts
     },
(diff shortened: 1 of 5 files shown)


chybisov (approved):

chybisov:
Thank you!

Issue #385, opened 2026-05-20
SolanaStepExecutor discards simulation details on TransactionSimulationFailed (code 1019)


antmaster2001:
## Problem

When `SolanaStepExecutor` runs `connection.simulateTransaction()` and the simulation fails, it throws a `TransactionError` with only the string `"Transaction simulation failed"` — discarding both the structured error object (`simulationResult.value.err`) and the program execution logs (`simulationResult.value.logs`).

This makes it impossible for SDK consumers to diagnose *why* the simulation failed (e.g. insufficient funds, program error, slippage, etc.) without re-simulating the transaction themselves.

## Current behaviour

```typescript
// SolanaStepExecutor.ts — current code
if (simulationResult.value.err) {
  throw new TransactionError(
    LiFiErrorCode.TransactionSimulationFailed,
    'Transaction simulation failed'  // ← no details
  );
}
```

The `TransactionError` constructor accepts an optional `cause?: Error`, but it is not passed here. The rich `simulationResult.value` is lost.

## What consumers see

```json
{
  "errorName": "SDKError",
  "errorMessage": "[TransactionError] Transaction simulation failed\nLI.FI SDK version: 3.16.3",
  "errorCode": 1019,
  "causeName": "TransactionError",
  "causeMessage": "Transaction simulation failed"
}
```

No actionable information to determine the root cause.

## Proposed fix

Attach the simulation details to the thrown error, either in the message or as a `cause`:

```typescript
if (simulationResult.value.err) {
  const simError = new Error(
    `Simulation error: ${JSON.stringify(simulationResult.value.err)}`,
    {
      cause: {
        err: simulationResult.value.err,
        logs: simulationResult.value.logs,
      }
    }
  );

  throw new TransactionError(
    LiFiErrorCode.TransactionSimulationFailed,
    `Transaction simulation failed: ${JSON.stringify(simulationResult.value.err)}`,
    simError
  );
}
```

This would let consumers inspect `.cause.cause.err` and `.cause.cause.logs` for the detailed Solana simulation result while keeping the top-level message human-readable.

## Environment

- SDK version: 3.16.3
- Chain: Solana (chainId 1151111081099710)
- Swap: USDC → SOL (same chain)

## Impact

Without simulation details, consumers must implement their own re-simulation as a workaround, which is unreliable since the transaction may have expired by the time the error is caught.

PR #386, opened 2026-05-25
fix: preserve Solana simulation details

samsamtrum:
## Summary
- include Solana simulation `err` and `logs` on the error cause when preflight simulation fails
- keep the existing human-readable `Transaction simulation failed: ...` message unchanged
- preserve bigint-safe JSON formatting for the top-level message

## Why
This addresses #385. Consumers currently receive code `1019` but lose the detailed Solana simulation result, so they have to re-simulate the transaction to diagnose the failure. Passing the details through the cause keeps the public message readable while making the original simulation payload inspectable.

## Testing
- `corepack pnpm install --frozen-lockfile --ignore-scripts`
- `corepack pnpm --filter @lifi/sdk build`
- `corepack pnpm --filter @lifi/sdk-provider-solana check:types`
- `corepack pnpm --dir packages/sdk-provider-solana test .unit.spec.ts --passWithNoTests`
- `corepack pnpm --filter @lifi/sdk-provider-solana build`
- `git diff --check`

### DIFF
--- a/packages/sdk-provider-solana/src/core/tasks/SolanaStandardWaitForTransactionTask.ts
+++ b/packages/sdk-provider-solana/src/core/tasks/SolanaStandardWaitForTransactionTask.ts
@@ -60,16 +60,23 @@ export class SolanaStandardWaitForTransactionTask extends BaseStepExecutionTask 
       )
 
       if (simulationResult.value.err) {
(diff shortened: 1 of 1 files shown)


chybisov:
Thank you! Addressed in #387

PR #438, opened 2026-07-28
fix(ethereum): skip Permit2 for signers with on-chain code

chmanie:
Permit2 verifies signatures from accounts with on-chain code via EIP-1271, so whether our plain ECDSA signature is accepted depends on the account implementation: Alchemy's ERC-6900 `SemiModularAccount7702` rejects it and the tx reverts before reaching the diamond, while MetaMask's `EIP7702StatelessDeleGator` recovers it fine. The signer is now probed with a read-only `isValidSignature` call before Permit2 is offered — accounts that reject it fall back to approve + execute, accounts that accept it keep the cheaper single-signature route, and accounts with no code skip the probe entirely.

Also fixes `getAccountCode` conflating "no code" with "RPC failed" — viem's `getCode` returns `undefined` for both — which was suppressing native EIP-2612 permits for every plain EOA.

Verified on Base against a 7702 account delegated to the same implementation as the report: [reverts with `0x151d90fe`](https://basescan.org/tx/0xb6afa8f9274d3994d03759db1c5d2c968695228490380c229113516c957a4109) before the fix, [succeeds via the diamond](https://basescan.org/tx/0x57f2cad5add5f43e5390a6399a1f4ecd433d0812073e7962bfd9eeba24fb1b13) after. The "mobile wallet browsers" framing in the report is incidental — the trigger is a delegated signer address, and it reproduces on desktop.

https://linear.app/lifi-linear/issue/JUMEMB-32/mobile-wallet-permit2-issue-transactions-fail-on-jumper-in-mobile


chybisov (changes requested):
Diagnosis looks right and the implementation is careful — memoizing a `Promise<boolean>` rather than a `boolean` is the right call, and the tests encode that reasoning. Checked out locally: 120 tests, `tsc --noEmit`, and `biome check` all clean.

Two things I verified so nobody repeats it: the Permit2 gate is complete (`signPermit2Message` is the only producer in the SDK, single call site), and re-enabling native permits leaves no redundant approval (`SetAllowanceTask.shouldRun` guards on `hasMatchingPermit`, which `EthereumNativePermitTask` sets first).

Four small things.

**1. The changeset should lead with the `getAccountCode` fix.** It has the wider blast radius: native EIP-2612 permits have been off for every plain EOA, and re-enabling them switches the live proxy entrypoint from `callDiamondWithPermit2` to `callDiamondWithEIP2612Signature` and drops the one-time `approve(permit2)`. Separately, "such accounts now fall back to the standard approve + execute path" only holds for non-permit tokens — `EthereumNativePermitTask.shouldRun` has no signer-code gate and `canAccountUseNativePermits` returns `true` for 7702, so those accounts land on the native permit path with `to: permit2Proxy`. That's fine (token-side `ecrecover`, no Permit2 verification), but the changeset should describe the route that actually runs.

**2. `EthereumStandardSignAndExecuteTask.ts:60`** — resolve the gate after the native-permit lookup, otherwise the `hasMatchingPermit` path pays an `eth_getCode` whose result is unused:

```ts
const signedNativePermitTypedData = signedTypedData.find(/* … */)
const permit2Supported = signedNativePermitTypedData
  ? false
  : await resolvePermit2Support(context, 'standard')
```

Also makes the two branches explicitly exclusive rather than implicitly so via `else if`.

**3. `checkPermitSupport`** — the only changed behavior with no test, and it's exported from `index.ts`. Two cases cover it: code-bearing owner → `permit2AllowanceSufficient: false` and `getAllowance` never called; plain EOA → allowance read still happens. The JSDoc ("2. Permit2 availability and allowance") no longer describes what it does now that it also encodes signer capability.

**4. `isSafeWallet`** — the `'0x'` short-circuit at lines 38-44 becomes reachable, so an undeployed/counterfactual Safe now returns `false` instead of falling through to the Safe API. The existing spec mocks `getAccountCode`, which is why the suite stays green through this. Worth one test plus a changeset line.

Not blocking: given "not verified end-to-end," I'd weight the device pass toward a plain-EOA EIP-2612 swap — that entrypoint now covers most users — then the two 7702 cases, permit-supporting and non-permit token.

chybisov (commented):
Looks good now, can we test it all end-to-end before the merge via preview?

chmanie:
Bad news and good news. After trying to reproduce this fix I saw that it was - in fact - not fixed. But it _can_ be reproduced outside of mobile wallets which I am attempting to do now using a preview build

chmanie:
@chybisov Tested within Jumper `develop`, looking good. Not only does this fix the issue with some Smart Wallets, it also fixes an issue where Permit2 transactions were used instead of just a Permit signature when the token actually supports EIP-2612, so it'll save users actual money ;)

chybisov (commented):
Ran a deep review on this. The probe genuinely fixes the JUMEMB-32 hard failure for wallets that revert on the raw-ECDSA `isValidSignature` probe — nice work. A few things should be addressed before merge, in order of severity:

1. **Relayed routes break on a false verdict** — `resolvePermit2Support.ts` only excludes the `batched` strategy, so a false probe verdict (including one caused by a single failed RPC call) moves the approval spender to `step.estimate.approvalAddress`. Scope, to be precise: relayed steps that settle via a native `Permit` never reach this choice — `EthereumCheckPermitsTask` signs the API's `Permit` typed data, sets `hasMatchingPermit`, and the allowance tasks skip. The only relayed state that does reach `SetAllowance` is the `PermitWitnessTransferFrom` one, where the relayer pulls funds via Permit2 and needs the token→Permit2 allowance — so the pull fails for missing allowance and the user paid for a useless approval. (Wallets holding an old MaxUint256 Permit2 allowance from the previous flow mask this; fresh wallets hit it.) Suggested fix: when the strategy is `relayed`, derive the spender from the step's typed data (`PermitWitnessTransferFrom` present → `fromChain.permit2`) and let the probe verdict steer only the standard flow.

2. **The probe's acceptance check is looser than Permit2's on-chain check** — `!!data && data.length >= 10` in `canAccountUsePermit2.ts` accepts wallets that return the ERC-1271 failure value `0xffffffff` (and fallbacks returning 4–31 bytes), so those accounts are classified Permit2-capable and still hard-fail on-chain — the same symptom this PR fixes, for a different wallet family. Minimal fix: require length `>= 66`. Better: decode and compare the returned `bytes4` against the magic value — switching the probe to `readContract` gives that nearly for free and matches the package's existing pattern.

3. **Revert-on-mismatch wallets get silently downgraded** — the probe signature deliberately recovers to a third party, so wallets that revert on signer mismatch (1/1 Safe `GS026`, `require(recovered == owner)` 7702 delegates, Sequence contract signatures) fail the probe but accept real signatures. They used Permit2 fine before this PR; now they pay an extra exact-amount approval tx per step. Worth either accepting explicitly (a comment would do) or handling, e.g. by verifying the real signature after signing.

4. **CCIP-Read follow on the probe** — the probe's `call` doesn't set `ccipRead: false`, so an `OffchainLookup` revert is auto-followed (outbound HTTP to contract-chosen URLs) and inverts the probe's own revert-vs-return rule. One-line fix on this call only (the shared client still needs CCIP for ENS).

Also worth a look, though not blocking: any RPC failure currently returns a memoized `false` with no retry or error signal, which silently downgrades healthy EOAs to a paid approval (and can double-approve on pause/resume). Happy to share the full findings list (15 items, incl. doc/test nits) if useful.

chybisov (approved):

PR #450, opened 2026-08-21
feat(stellar): require the account reserve before spending XLM

melianessa:

None yet — this came out of a reported mainnet swap failure. Happy to attach one.

A Stellar account may not drop below its minimum balance:

```
(2 + numSubEntries + numSponsoring - numSponsored) x baseReserve + native selling liabilities
```

Spending into that floor makes the native Stellar Asset Contract reject the `transfer` with `Error(Contract, #10)` (`BalanceError`). Because it fails inside a Soroban invocation, the user sees an opaque simulation `HostError` with a 20-event diagnostic dump rather than a balance error — the reported case was a max XLM swap that left 200 stroops against a 1.5 XLM floor.

Nothing in the SDK modelled this. `getStellarBalance` reads the raw SAC `balance()`, which for native XLM includes the reserve, so `checkBalance` saw a sufficient balance and passed. Worse, its final-attempt slippage rescue rewrote `fromAmount` to `have - reserved`, i.e. deliberately drained the account to zero — manufacturing the failure even from a safe user input.

**Why an optional `SDKProvider` method.** `getNativeReserve?(client, walletAddress)` is opt-in, so chains without an unspendable floor are untouched (no extra call, no behavior change) and any future chain with one — Solana's rent-exempt minimum is the same shape — implements the same hook. The alternative, a `CheckBalanceOptions` field plus a `StellarCheckBalanceTask` subclass, needed a new task per ecosystem and left the generic `CheckBalanceTask` unaware of the reserve, so the slippage rescue would still have trimmed into it.

**Why it folds into `reservedOverhead`.** That one helper feeds the sufficiency comparison, the "never trim overhead by slippage" rule, and the rescue's floor. Adding `reservePart` there fixes all three at once: the shortfall is now reported up front, and the rescue trims to `balance - gas - fees - reserve` instead of to zero. The reserve is read alongside the balances inside the retry loop, so it shares the existing 6-attempt backoff and both `BalanceError` paths, and it is only read when a step actually spends the native token.

**Why a flat trustline headroom.** `LiFi::internal_swap` calls `try_trust(&sender)` on the destination token (CAP-73) before swapping, so a route into an asset the sender does not yet hold **adds a subentry inside the transaction being checked** — the floor at execution time is one base reserve above the floor read beforehand. The reported account has `subentry_count: 0`, so its at-rest floor is 1.0 XLM while its simulation reported 1.5 XLM; without headroom a Max amount would leave 1.0 XLM and still fail on any first-time destination token. Which token a route ends in is not known at the point of the read, and over-reserving 0.5 XLM only idles balance whereas under-reserving fails the transaction, so the headroom is unconditional. An exact answer would need the destination token threaded into the read plus a trustline probe per quote; a follow-up option is for the API to report it, since the backend already resolves trustline status.

`baseReserve` is pinned at 0.5 XLM rather than read per call: it lives in the ledger header, which Stellar RPC does not expose, and changing it requires a validator-voted network upgrade (unchanged since 2015).

No UI. Verified against mainnet with the account from the report — the computed reserve is `15000000` stroops, byte-identical to the floor its own failing simulation named:

```
["resulting balance is not within the allowed range", 15000000, 200, 9223372036854775807]
```

Cross-checked against Horizon: `subentry_count: 0`, no sponsorships, no selling liabilities.

The API change is one optional method on `SDKProvider`, which only matters to someone implementing a custom provider — leaving the docs box for a reviewer to call.

**Tests:** 14 new unit tests — 8 covering the reserve computation (bare account, per-subentry, sponsorships both directions, selling liabilities, negative clamp, missing account, read failure) against real `AccountEntry` XDR, and 6 covering `checkBalance` (native shortfall with the reserve message, sufficient balance, slippage trimming that preserves exactly the reserve, reserve required on top of gas when the source token is not native, no read when no native requirement exists, and read failure).

### DIFF
--- a/.changeset/stellar-native-reserve.md
+++ b/.changeset/stellar-native-reserve.md
@@ -1,0 +1,8 @@
+---
+"@lifi/sdk-provider-stellar": minor
+"@lifi/sdk": minor
+---
+
+Account for the Stellar native reserve before spending XLM. `SDKProvider` gains an optional `getNativeReserve`, which the Stellar provider implements by reading the sender's account entry and computing `(2 + numSubEntries + numSponsoring - numSponsored) x baseReserve + sellingLiabilities`, plus one base reserve of headroom for the trustline `LiFi::internal_swap` opens on the sender's behalf (`try_trust`) inside the transaction being checked.
+
+`checkBalance` now requires that reserve on top of the source amount, gas and non-included fees whenever a step spends the chain's native token, and preserves it when trimming the source amount by slippage. Previously a swap of nearly the whole XLM balance passed every pre-flight check and then failed in simulation with an opaque `HostError: Error(Contract, #10)` from the native Stellar Asset Contract; it now fails fast with a balance error naming the reserve, and the slippage rescue no longer proposes an amount that drains the account below it.
--- a/packages/sdk-provider-stellar/src/StellarProvider.ts
+++ b/packages/sdk-provider-stellar/src/StellarProvider.ts
@@ -9,6 +9,7 @@ import {
   type TokenAmount,
 } from '@lifi/sdk'
 import { Networks, StrKey } from '@stellar/stellar-sdk'
+import { getStellarAccountReserve } from './actions/getStellarAccountReserve.js'
 import { getStellarBalance } from './actions/getStellarBalance.js'
 import { resolveStellarAddress } from './actions/resolveStellarAddress.js'
 import { StellarStepExecutor } from './core/StellarStepExecutor.js'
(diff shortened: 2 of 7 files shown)


melianessa:
Closing in favour of a backend-side design.

The reserve is better computed server-side: the backend already reads the sender's account entry while building the envelope (`stellar.invocation.builder.ts` fetches it for the sequence number, and `StellarAccount` already carries `subentryCount`, `numSponsoring`, `numSponsored`, `nativeBalance`), so the check costs no extra RPC there, and only the backend knows the destination token — so it can consult trustline status instead of assuming a flat headroom for the trustline `LiFi::internal_swap` opens.

It will land as: the reserve exposed on the gas-suggestion endpoint so clients can size a Max button, the quote clamped to a sendable amount so the quoted `toAmount` is what the user actually receives, and a guard at envelope build time to replace the opaque `HostError: Error(Contract, #10)` with a readable error. That also avoids the part of this PR that was hardest to justify — `checkBalance` silently trimming `fromAmount` after the quote had already been shown.

Diagnosis and mainnet verification are preserved in the description above for whoever picks up the backend work.

PR #452, opened 2026-08-24
fix(ethereum): probe EIP-1271 before offering a native EIP-2612 permit

chmanie:
Tokens whose `permit` verifies through a `SignatureChecker` — Circle's USDC among them — branch on `owner.code.length` just as Permit2 does, so a code-bearing owner is verified via EIP-1271 and strict delegates reject the bare ECDSA signature with `EIP2612: invalid signature`. Code-bearing accounts are now probed with the same `isValidSignature` check already used for Permit2, rather than passed on the grounds that they can sign ECDSA.

Follow-up to #438, which fixed the Permit2 half and left this one. Verified on Arbitrum against the reported account `0x4471467c…9c99`: both gates now decline it, while a MetaMask 7702 account and a plain EOA keep both fast paths.

https://linear.app/lifi-linear/issue/JUMEMB-32/mobile-wallet-permit2-issue-transactions-fail-on-jumper-in-mobile


chmanie:
Tested the fix and works properly now.

chybisov (commented):
Right diagnosis, and asking the token's question instead of the signer's is the correct reframe. The gate just widened further than the fix needs.

`acceptsRawEcdsaSignature` now serves two callers with incompatible requirements. `encodePermit2Data` forwards the signature as `bytes`, so any length works. `encodeNativePermitData` calls `parseSignature` and needs exactly 65 bytes. A wallet that returns a failure selector rather than reverting passes the probe, since a `bytes4` pads to a full word and `data.length` is 66, which `canAccountUsePermit2.unit.spec.ts` already relies on. Solady's `ERC1271` does exactly that by design. Such an account is now offered a native permit, `hasMatchingPermit` skips the approval, and the step throws in `parseSignature` after the user has signed. Safe reverts on the probe and stays blocked, so the PR description's reasoning holds for Safe but not for the group whose verdict changed.

Putting the old shape check in front of the probe fixes it. `code === '0x' || isDelegationDesignatorCode(code)` is exactly `!isSmartContractWalletCode(code)`, so 7702 delegates are still probed and nothing the PR set out to fix is lost.

```ts
if (code === '0x') {
  return true
}
if (!isDelegationDesignatorCode(code)) {
  return false
}
return acceptsRawEcdsaSignature(client, chainId, address)
```

Keep the `code === undefined` return above it. Two follow-ons. The JSDoc and the changeset both currently say accounts with code are probed, which stops being true. And `blocks a smart-contract wallet that rejects raw ECDSA` forces the probe to `false`, so it passes whatever the gate does. Drop that mock, since `beforeEach` already defaults the probe to `true`, and assert the probe was never called.

I checked whether the encode boundary needs its own guard and it does not. `encodeNativePermitData` has one call site, and the only other producer of a `Permit` in `signedTypedData` is `EthereumCheckPermitsTask`, whose `shouldRun` predicate is the same one `isRelayerStep` uses, so those steps route to the relayed executor instead. This gate is the sole live entry point.

Non-blocking, happy for these to be separate. `checkPermitSupport` reports no native permit support when no wallet is connected, because the public client fallback has no account and the new guard turns that into `false`. Threading `ownerAddress` and `chainId` into the gate fixes that and the `viemClient.chain?.id` read together. The probe is also neither deduped nor memoized, so one `checkPermitSupport` call can issue two of them. And the new module has no spec of its own, since both existing specs reach it only indirectly.

chmanie:
Okay tested
* Alchemy EIP-7702
* Plain EOA
* MetaMask EIP-7702

all work as expected

chybisov (approved):

PR #464, opened 2026-09-03
feat(permits): export generic Permit2 builders for arbitrary spenders

Kayanski:
The `SignatureTransfer` and `AllowanceTransfer` EIP-712 builders in `@lifi/sdk-provider-ethereum` already accept an explicit `spender`, but only the LI.FI-proxy-bound wrappers (`getNativePermit`, the internal `getPermitTransferFromValues` which hardcodes `spender = chain.permit2Proxy`) are reachable from the public entrypoint.

This surfaces the existing generic builders publicly, namespaced to mirror [`@uniswap/permit2-sdk`](https://github.com/Uniswap/permit2-sdk) (`SignatureTransfer` / `AllowanceTransfer`), so integrators can sign Permit2 messages for **any** spender — their own contract, Uniswap's Universal Router, etc. — without routing through the LI.FI Permit2 proxy / diamond.

### Usage

```ts
import { SignatureTransfer } from "@lifi/sdk-provider-ethereum";

const permit = {
  permitted: { token, amount },
  spender,            // any spender, not just chain.permit2Proxy
  nonce,
  deadline,
};
const { domain, types, message } = SignatureTransfer.getPermitData(permit, permit2Address, chainId);
// walletClient.signTypedData({ domain, types, message, primaryType: "PermitTransferFrom" })
```

### Notes / scope

- **No behaviour change** — pure re-export of existing pure modules (`permits/signatureTransfer.ts`, `permits/allowanceTransfer.ts`), plus `permit2Domain`.
- `export * as` is banned by the repo's `noReExportAll` biome rule, so the namespaces are built via `import * as …` + a named `export { … }` (same shape Uniswap uses).
- Adds `signatureTransfer.unit.spec.ts` asserting a message can be built and hashed for a **non-proxy** spender, and that a different spender yields a different digest.
- Full `pnpm check && check:types && check:circular-deps` pass (run by the commit hook); `sdk-provider-ethereum` unit tests green (32 passed).

Follow-up (out of scope here): wiring a proxy-less Permit2 execution path into the SDK's step executor for non-LI.FI spenders.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/sdk-provider-ethereum/src/index.ts
+++ b/packages/sdk-provider-ethereum/src/index.ts
@@ -1,3 +1,11 @@
+// Generic, spender-agnostic Permit2 EIP-712 builders, namespaced to mirror
+// @uniswap/permit2-sdk. Lets integrators sign Permit2 messages for any spender
+// (their own contract, Uniswap's routers, ...) without routing through LI.FI's
+// Permit2 proxy. Exposed as namespace objects to avoid the duplicate
+// getPermitData/hash names across the two modes.
+import * as AllowanceTransfer from './permits/allowanceTransfer.js'
+import * as SignatureTransfer from './permits/signatureTransfer.js'
+
 // biome-ignore lint/performance/noBarrelFile: module entrypoint
 export { checkPermitSupport } from './actions/checkPermitSupport.js'
 export {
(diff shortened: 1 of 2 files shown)


Kayanski:
Closing — not needed for the use case that prompted it. The Uniswap-via-LI.FI single-step permit flow builds the Permit2 typed data on the backend and signs it generically in the SDK (#465), so the exported builders here aren't required. Leaving this as a possible standalone DX improvement for later if there's interest, but no immediate consumer.

PR #475, opened 2026-09-10
feat(ethereum): classify typed data by lane and support caller-supplied Permit2 intents

chybisov:

[JUMADV-90 — Widget: native Permit2 (PermitSingle) signing for custom swap providers](https://linear.app/lifi-linear/issue/JUMADV-90/widget-native-permit2-permitsingle-signing-for-custom-swap-providers)

Supersedes #465. Companion: [lifinance/types#577](https://github.com/lifinance/types/pull/577), released as `@lifi/types@18.6.0`.

### The problem

A step's `typedData` was read by three overlapping boolean sniffs at six decision points. All three asked *"does this step have typed data?"* and none asked *"what kind?"*. A caller-supplied Permit2 intent — a `PermitSingle` naming a third-party spender, which the user signs and the API embeds into the calldata — was indistinguishable from a gasless relayer intent.

### Why not #465's `hasNonNativePermit`

Reviewing that PR surfaced two blockers, both from one cause: a predicate over a *step*, answering a question that only makes sense per *entry*.

1. It returns true for `PermitWitnessTransferFrom`, so gasless steps lose the Permit2 spender.
2. `Order` steps route to `standard` and fail with `TransactionUnprepared`, breaking CowSwap, 1inch Fusion and Velora Delta.

### The approach

One classifier, `getTypedDataLane`, classifies **one entry**:

| Lane | Meaning |
| --- | --- |
| `native-permit` | EIP-2612 `Permit`. Stands in for the ERC-20 allowance. |
| `caller-intent` | `PermitSingle` for a third-party spender. The user signs it, the API embeds the signature, and the SDK sends that calldata **untouched** to its own target. |
| `relayer-intent` | `PermitWitnessTransferFrom`, or a message whose spender is `chain.permit2`. A relayer pulls the funds and pays the gas. |

Step-level questions are folds over the entries. The lanes are **not** mutually exclusive, and one named predicate carries that: `isCallerIntentLane` = has a caller intent **and** no relayer intent. Every gate that must leave a caller-executed step alone asks exactly that, in one spelling.

**Rule order is load-bearing.** A native `Permit` is decided *before* the gasless rule — see the regression note below. `Order`, `PermitBatch`, `Agent`, `OrderCancellations`, `SignedBatchedCall` and every future primary type fall to the `relayer-intent` default, so their behaviour is unchanged.

### Verified against the real producers, not assumed

Three backends were read to settle questions the SDK cannot answer alone:

- **`Order` messages carry no `spender` field.** CoW emits 12 fields, 1inch 8, Velora 14. So the spender rule cannot misfire on an order-based tool.
- **The gasless relayer emits an EIP-2612 `Permit` whose `spender` IS the canonical Permit2** (`gasless/src/signature/payload.ts:78`), only when the allowance is short, always paired with a witness entry.
- **The first real `PermitSingle` producer** (jumper-backend #1216, Fynd) emits `type: 'lifi'`, `typedData: [permit]`, `message.spender` = its router, and `domain.verifyingContract` = `estimate.approvalAddress` = Permit2 — which is exactly the shape the MAX-approval rule below was written for.

### A regression this PR found in itself

An earlier revision of this branch classified the gasless native permit as `relayer-intent`, because its spender is Permit2. That made `EthereumCheckPermitsTask.shouldRun` false, so `hasMatchingPermit` was never set, so `EthereumSetAllowanceTask.shouldRun` became true — **a gasless user would have been asked to fund an on-chain approval**, in exactly the case gasless emits that permit.

Fixed in `891d85cb` by deciding native permits first. A mechanical check over 19 primary types × 5 spender shapes × 2 chains — 190 inputs — confirms exactly one cell moves, and it moves back to what `main` does.

### Behaviour changes

- A new `EthereumSignStepIntentTask` signs caller intents, after the allowance tasks and before prepare. It reuses the `PERMIT` action, so no widget change is needed. Both widget repos were read to confirm this.
- The token approval to Permit2 is unlimited only when `estimate.approvalAddress` matches the `domain.verifyingContract` of every caller intent, and never when `disableMessageSigning` stops the SDK from signing it.
- The outer-transaction gas check is enforced for caller-intent steps — the user funds those.
- `EthereumPrepareTransactionTask` no longer erases a caller intent when the API answers `typedData: []`. That erasure survived a retry, and an EIP-7702 `atomicityNotReady` retry needs no user action, so the SDK would have wrapped the caller's calldata and retargeted it to `permit2Proxy`.

### Routing: two additional relayer signals, and one narrowing

Three late changes, driven by reading the live backends rather than the diff.

**Narrowed the relayer test back toward `main`.** The branch had stopped force-routing *every* typed-data step; the feature only ever needed the caller-intent lane carved out. The wider rule re-opened a window that caused a production incident (JUMEMB-88): a `[Permit]`-only step could resolve `batched`, queue an approve into a batch, then flip to `relayed` at prepare and drop it. The rule is now:

```ts
if (step.executionType === 'message' ||
    (!!step.typedData?.length && !isCallerIntentLane(step, fromChain))) {
  return 'relayed'
}
```

Every typed-data shape returns to `main`'s behaviour except the one this feature exists for.

**`executionType === 'message'` as an OR, never a replacement.** lifi-backend sets it on every Hyperliquid flow at routes time, while `typedData` only arrives at `/advanced/stepTransaction` — the window JUMEMB-102 describes. It is an OR because the gasless lane reports `executionType: 'transaction'` **by design** (lifi-backend's schema rejects `gasless: true` with `executionType: 'message'`), so gasless depends entirely on the typed-data inference. jumper-backend sets the field nowhere at all, which is the other reason it cannot be the primary rule.

**A structural rule at dispatch time.** A step carrying typed data with no `transactionRequest` has no viable path — `batched` requires a transaction, `standard` throws without one, and only `relayed` needs neither. It is decided in `EthereumSignAndExecuteTask`, not in the strategy resolver, because before prepare a caller-intent step that *will* receive a transaction and one that never will are indistinguishable.

**And a defect that combination exposed:** the relayed task's "already signed" check only recognised `primaryType === 'Permit'`, so a caller intent signed before prepare was signed again by the relayer. One message, two wallet prompts. Fixed.

Not applicable — SDK execution-path logic, no UI surface.

## Testing

- **273 tests** in `@lifi/sdk-provider-ethereum`, **970** across all 7 packages.
- 13 new spec files, including four tasks that previously had none — among them `EthereumRelayedSignAndExecuteTask`, the shipped gasless path.
- An exhaustiveness table pins every `TypedDataPrimaryType` to a lane, so a new primary type fails the suite until someone assigns it deliberately.
- Two prohibitions are pinned by tests because reviewers proposed breaking them: `getUpdatedStep` must not use `hasRelayerIntent` (it would break CowSwap retries), and the intent task must emit `ACTION_REQUIRED`, not `MESSAGE_REQUIRED` (the widget has no mapping for the latter and would render a blank row).
- Mutation-tested: 25 mutations, all surviving mutants since closed.

Unit-tested only. End-to-end coverage needs the Fynd producer in jumper-backend#1216.

## Release shape

`@lifi/sdk-provider-ethereum` takes a **minor**, `@lifi/sdk` a **minor** for the `@lifi/types` bump to `^18.6.0`.

**`isRelayerStep` is removed.** It only told you the step carried typed data, which no longer says how a step executes: a caller-supplied Permit2 intent is signed inline and sent by the user. It had no callers left inside the SDK. Use `isGaslessStep` to ask whether the relayer pays the gas.

Shipping that under a minor is a deliberate maintainer decision. Both widget repos hold zero references, and `widget-provider-ethereum/src/index.ts` is an explicit export list with no `export *`, so nothing is transitively re-exported — the exposure is third-party integrators importing the symbol directly. The changeset announces the removal so it appears in the changelog.

## Verified against the execution-flow specs

#478 pins `main`'s behaviour across 14 EVM flows. Combining this branch with it gives **329 pass, 0 fail**, with biome, tsc, madge and knip clean.

Two of those specs changed, because the behaviour they pinned genuinely moved:

| Change | Effect |
| --- | --- |
| A signature-only caller intent signs under its own `PERMIT` action | It signed under `SWAP` before, and the action list now carries a second, durable entry. The widget needs copy for `PERMIT` + `ACTION_REQUIRED` before rollout. |
| That step no longer raises `MESSAGE_REQUIRED` | `EthereumSignStepIntentTask` has already signed it, so the relayed task finds nothing left to sign. |

Everything else — gasless, orders, Hyperliquid, native permits, the SDK's Permit2 flow, batched approvals, reset-then-approve, cross-chain, the 7702 retry, paused execution — passes unchanged.

The suite earned its place twice. It caught the double-signature defect on its first run, and then a defect this branch had introduced.

### The lane that signs and the lane that waits

An earlier revision decided the relayed lane at the dispatch site, in `EthereumSignAndExecuteTask`. That routed the signature correctly but left the strategy stored in the context on `standard`, and `EthereumWaitForTransactionTask` reads that stored value — so it waited through `waitForTransactionReceipt` with `action.txHash` undefined. A relayed step carries a `taskId` and no hash. jumper-backend's `uniswap-trading` permit step is exactly that shape.

`getEthereumExecutionStrategy` now owns the decision alone. The rule it gained is bound to the single call from `EthereumPrepareTransactionTask`: after the re-quote, typed data with no transaction request can only execute through the relayer. Before prepare the same test is unsound — a caller intent that receives its transaction from `/stepTransaction` is indistinguishable from one that never will, and calling it relayed there costs the step its EIP-5792 batching.

Two new flow specs pin both halves. They need #478's harness and assert this branch's behaviour, so they can land on neither branch today. They are parked on the local branch `wip/jumadv-90-flow-specs`, to be cherry-picked after the rebase.

Three smaller corrections ride along:

- `signTypedDataEntries` announces the signing status once per round again. Relay from Hyperliquid carries two messages, and the extracted loop had started announcing each one.
- `isGaslessTypedData` compares the spender against `chain.permit2` without case, matching `getApprovalAmount`.
- `preserveCallerIntents` documents the case it actually guards. No backend emits `typedData: []` — jumper-backend drops the key, lifi-backend converts it to `undefined`, and the gasless service rejects it at the schema.

## Known items, deliberately not fixed here

1. **`uniswap-trading`**, merged to jumper-backend develop, attaches its `PermitSingle` to a dedicated step that never receives a `transactionRequest`. It keeps working with no backend change. Declaring `executionType: 'message'` on that step would make the verdict correct earlier and skip a needless batching probe — worth doing, not required.
2. **The stranded-approval hazard** (JUMEMB-102) is narrowed but not closed: `executionType` covers Hyperliquid, and the narrowing removes the window this branch had opened, but gasless and every jumper flow still rely on the typed-data inference.

## Public surface

The classifier and all new helpers are internal, absent from both `index.ts` files. `isGaslessStep` keeps its signature, behaviour and export — the widget depends on it in three places.

The third item is unchecked deliberately: the only public-surface change is the removal of `isRelayerStep`, which the changeset announces. Point me at a `public-docs` page that lists it and I will update it.


chybisov:
@Kayanski could you review this one, and give the preview a spin against uniswap-trading and Fynd?

This replaces the approach in #465: instead of a single `isRelayerStep` check, the SDK now classifies every `typedData` entry into a lane and picks the execution strategy from that. It means the backend no longer needs the approve step, the capture-only relayer step or the status poll — but it does mean the step has to be shaped so the SDK reads the lane you intend. Writing that contract out here so you can adapt both integrations without reverse-engineering it.

## Preview

```
@lifi/sdk@0.0.0-preview-99c40e1b
@lifi/sdk-provider-ethereum@0.0.0-preview-99c40e1b
```

`@lifi/sdk` and every `@lifi/sdk-provider-*` must move together — same version, one copy in the tree.

Widget build with the preview pinned, for the UI side: jumperexchange/jumper-widget#121 → https://pr-121.widget-preview-6pd.pages.dev

## How the SDK classifies a step

`getTypedDataLane.ts` reads **each** `typedData` entry, in this order:

| Test | Lane |
| --- | --- |
| `primaryType === 'Permit'` | native-permit |
| `primaryType === 'PermitWitnessTransferFrom'`, or `message.spender` equals `chain.permit2` | relayer-intent |
| `primaryType === 'PermitSingle'` | caller-intent |
| anything else | relayer-intent |

A step is the **caller lane** only when it has at least one caller-intent entry and **zero** relayer-intent entries (`isCallerIntentLane`).

`getEthereumExecutionStrategy.ts` then picks the strategy:

- `step.executionType === 'message'` → `relayed`. This wins over everything else, and it is the only lane signal that reaches the allowance tasks, which run before a `transactionRequest` can exist.
- typed data present and not the caller lane → `relayed`.
- after prepare: typed data present and no `transactionRequest` → `relayed`.
- otherwise → `batched` or `standard`.

## What a caller-lane step must look like

1. Do **not** set `executionType: 'message'`. It forces the relayer lane before anything else is considered.
2. `primaryType: 'PermitSingle'`.
3. `message.spender` is the integrator's router (Universal Router / Fynd), **not** the Permit2 address. A spender equal to `chain.permit2` classifies the entry as a relayer intent.
4. `estimate.approvalAddress` is the canonical Permit2 address.
5. `domain.verifyingContract` is that same Permit2 address. `getApprovalAmount.ts` approves `MaxUint256` only when every caller-intent entry's `verifyingContract` matches `approvalAddress`; otherwise the user re-approves `fromAmount` on every swap.
6. `/advanced/stepTransaction` returns a `transactionRequest`. The user sends it.

## The `/advanced/stepTransaction` body changed

`getUpdatedStep.ts` (`getStandardUpdatedStep`) strips `step.typedData` and sends the signed entries in its place:

```jsonc
POST /advanced/stepTransaction
{ ...step, "typedData": [ { /* the same entry */, "signature": "0x…" } ] }
```

So the request schema needs to accept a `signature: Hex` on each entry, and the handler builds the calldata from it.

## What you return

`preserveCallerIntents.ts` keeps the caller's intent when the response omits `typedData`, carries only native permits, or carries `[]`. All three are safe.

What is not safe: returning a relayer-intent entry on a caller-lane step. A retry re-reads the lane from the answer, flips the step to the relayer, and the user never gets a transaction.

## Order of operations

`EthereumStepExecutor.ts:141-149` — allowance → **sign the `PermitSingle`** → prepare → send.

- The user approves Permit2 first, then signs, then `/advanced/stepTransaction` runs.
- Read the Permit2 nonce from `allowance(owner, token, spender)` at quote time.
- `sigDeadline` and `expiration` need to survive approve + sign + prepare + send, not just the quote.

## If uniswap-trading stays on the relayer lane

Set `executionType: 'message'` on the step. Without it the SDK cannot see the lane before the allowance tasks run, and the user pays for an approval the relayer does not need.

My recommendation is the other direction: move uniswap-trading to the single-step Fynd shape — one swap step with `typedData` attached and `approvalAddress: info.permit2Address` — and drop `buildApproveSteps`, `buildPermit2Step`, `advancedRelay` and `relayerStatus`.

## Two open questions

1. **`disableMessageSigning`.** When an integrator sets it, `EthereumSignStepIntentTask` does not run, but `isCallerIntentLane` still reports the caller lane, so the backend gets no signature — and Fynd throws without one. Either the SDK drops caller-intent steps under that flag, or the backend needs a path that builds the transaction without a signature. I lean towards handling it in the SDK; tell me if that breaks anything on your side.
2. **Nonce freshness on retry.** The signature is produced before `/advanced/stepTransaction`, so the nonce comes from the quote. If a user sits on a quote through an approval, is your nonce still good, or should the SDK re-quote before signing?

Nothing here is settled — if a rule makes your integration awkward, say so and I will move the SDK rather than have you work around it.

packages/sdk/src/core/tasks/helpers/waitForTransactionStatus.ts
```diff
@@ -38,9 +39,14 @@ export async function waitForTransactionStatus(
                   statusResponse.status,
                   statusResponse.substatus
                 ),
-              txLink: (statusResponse as FullStatusData).bridgeExplorerLink,
+              // Most bridges implement no `getExplorerLink`, so fall back to
+              // the LI.FI explorer rather than leave the user with no link.
+              txLink:
```
Kayanski, line 44: This is just wonderful

packages/sdk-provider-ethereum/src/core/tasks/EthereumSignStepIntentTask.ts
```diff
@@ -0,0 +1,49 @@
+import { BaseStepExecutionTask, type TaskResult } from '@lifi/sdk'
+import type { EthereumStepExecutorContext } from '../../types.js'
+import {
+  getTypedDataInLane,
+  isCallerIntentLane,
+} from '../../utils/getTypedDataLane.js'
+import { signTypedDataEntries } from './helpers/signTypedDataEntries.js'
+
+/** Signs the Permit2 messages a caller attached to its own step, before prepare. */
+export class EthereumSignStepIntentTask extends BaseStepExecutionTask {
```
Kayanski: The naming here is not ideal, can we rename to 
EthereumSignPermit2Task ? or something that has Permit2 in the name if it's the purpose of it ?

Kayanski (approved):

